PR #3, opened 2019-04-15
Add sessionToken field to Credentials, and add it to request headers

paulreimer:
Add `sessionToken` field to Credentials, and add it to (unsigned) request headers as `X-Amz-Security-Token`.


paulreimer:
Only thing to notice here, apparently some AWS services require the `X-Amz-Security-Token` to be included in the `signedHeaders`, but other services (like IoT as I am using) require it NOT in the `signedHeaders`.

/shrug

PR #8, opened 2019-06-25
add create queue api

ipconfiger:
Implement this API：
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html

also add unit test support
The new method have been tested and works.

finally, all keys, account info been all cleared.


pubspec.yaml
```diff
@@ -12,3 +12,6 @@ dependencies:
   crypto: ^2.0.1
   http_client: ^0.5.0
   xml: ^3.0.0
+
+dev_dependencies:
+  test: '>=0.12.30 <1.4.0'
```
isoos: please add a newline

isoos: why not just use `^1.3.0`?

lib/sqs.dart
```diff
@@ -27,6 +27,32 @@ class Sqs {
   /// Returns a new SQS queue, inheriting the properties of this instance.
   SqsQueue queue(String queueUrl) => new SqsQueue(queueUrl,
       credentials: _credentials, httpClient: _httpClient);
+
+  /// Create a new SQS queue
+  Future<SqsQueue> createNew({endpoint:null, queueName:null, maxSize:'1024', retentionPeriod:'345600'}) async {
```
isoos: Please add types to the parameters.

test/impl_test.dart
```diff
@@ -0,0 +1,31 @@
+library aws_client.test.impl_test;
+
+import "package:test/test.dart";
+import 'package:aws_client/aws_client.dart';
+import 'package:aws_client/sqs.dart';
+import 'package:http_client/console.dart';
+
+void main() {
+
+  group('SQS-queue', (){
```
isoos: Please update the test to read accessKey and secretKey from environment variables (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`) (maybe add that detection to `Credentials`?).

lib/sqs.dart
isoos: I think we should also use `package:meta` here, and add `@required String endpoint` and `@required String queueName` (without and `:null` initialization which is implicit).

isoos: I think maybe `create` would be enough for a method name.

isoos:
Thanks for the code, I think this is solid, I have only a few nitpicks that I've commented. Please address them and I will merge it.

ipconfiger:
All issues fixed, and i change parameter endpoint to region, because it's easy to add a wrong endpoint url, if miss the last '/', the signature won't be correct.

isoos:
@ipconfiger Thanks! Let me know if you want to add more things or if I should publish it as it is now.

PR #10, opened 2019-06-26
Add SNS Service API support

ipconfiger:
Implemented APIs:
1. create new endpoint
2. create new topic
3. subscribe topic
4. unsubscribe topic
5. push message to endpoint
6. push message to topic


lib/sns.dart
```diff
@@ -0,0 +1,228 @@
...
+    assert(this._credentials != null);
+    assert(this._httpClient != null);
+    assert(this._region != null);
+  }
+  final Credentials _credentials;
+  final http.Client _httpClient;
+  final String _region;
+
+  /// execute real request
+  Future<String> executor(Map<String, String> parameters) async {
```
isoos: What does this method?

lib/sns.dart
```diff
@@ -0,0 +1,228 @@
...
+  final RequestExecutor _executor;
+  final Credentials _credentials;
+  final http.Client _httpClient;
+  final String _arn;
+  final String _region;
+
+  /// push message to this Topic
+  /// implements of https://docs.aws.amazon.com/sns/latest/api/API_Publish.html
+  /// return MessageId type:String
+  Future<String> pushNotification(String body) async {
```
isoos: Is this the same as `SnsEndpoint.pushNotification`? If so, maybe it can be inherited from that class (making endpoint the superclass of topic)?

ipconfiger: Thers's two pushNotification target, the Topic and the Endpoint, the method in SnsEndpoint will send notification to the Endpoint, the method in SnsTopic will send notification to the Topic.

lib/sns.dart
ipconfiger: This method build the AwsResponse, execute the query and get de response. en, so there's no need to pass the tree property: credentials, httpClient, region to next Class every time. I forget the remove then this classes, i will fix later...

isoos: In that case, I think this should be private method to this class, e.g. rename it to `_sendRequest`?

ipconfiger: en, sounds nice, i'm no good at naming, pls wait a moment.

ipconfiger: fixed that

lib/aws_client.dart
```diff
@@ -26,4 +29,9 @@ class Aws {
   /// Returns an SQS service, inheriting the properties of this instance.
   Sqs get sqs =>
       _sqs ??= new Sqs(credentials: _credentials, httpClient: _httpClient);
+
+  /// Returns an SNS service, inheriting the properties of this instance.
+  Sns sns(String region) =>
+      _sns ??= new Sns(credentials: _credentials, httpClient: _httpClient, region: region);
```
isoos: This is probably wrong, because if I call that with a different `region`, I will be served with an instance that was created earlier. Maybe we could have an `final _snsMap = <String, Sns>{};` and here `return _snsMap.putIfAbsent(region, () => new Sns(...));`?

isoos: Or maybe just return the `new Sns(...)`

ipconfiger: Generally, in most 80% case, only one region needed, so we have to pass region to every next actions, that's too  cumbersome. If more than one region we use,  pass different Sns instance for each region even make sense. Right?

isoos: Let's use `return new Sns(...)` with region. It is cheap, and people can still store the sns instance separately if needed. But current initialization with "first region wins and you have no chance to use another one" is not something we want to commit.

ipconfiger: en, current initialization used like:

    final sns1 = aws.sns('region one'); // this return a new Sns instance with region : region one
    final sns2 = aws.sns(region two'); // this return a new Sns instance with region : region two

Just like what you said: use return new Sns(...) with region.  or what i missed?

ipconfiger: Ok, i see what i missed, i will fix that a minute later

isoos:
Thanks for the PR! Do you want me to publish it or do you have more stuff to come?

isoos:
FYI: I've updated a few things in analysis options and in the code, and published a new version.

PR #14, opened 2019-11-15
Add generated services

Schwusch:
I've made an AWS SDK generator and I thought I'd show one of the services it generates for you to give your thoughts on. If there is interest, I can provide the other AWS services as well.

### DIFF
--- a/examples/lambda.example.dart
+++ b/examples/lambda.example.dart
@@ -4,11 +4,10 @@
 
 import 'dart:convert';
 import 'dart:io';
+import 'package:aws_client/aws_client.dart';
+import 'package:aws_client/lambda.dart';
+import 'package:aws_client/src/credentials.dart';
 import 'package:http_client/console.dart';
-
-import '../lib/aws_client.dart';
-import '../lib/lambda.dart';
-import '../lib/src/credentials.dart';
 
 main(List<String> args) async {
   final Map<String, String> environment = Platform.environment;
--- a/lib/aws_client.dart
+++ b/lib/aws_client.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be found
 // in the LICENSE file.
 
+import 'package:aws_client/dynamodb/dynamodb-2012-08-10.dart';
 import 'package:http_client/http_client.dart' as http;
 
 import 'lambda.dart';
@@ -27,12 +28,14 @@ class Aws {
 
   /// Returns an SQS service, inheriting the properties of this instance.
   Sqs get sqs =>
-      _sqs ??= new Sqs(credentials: _credentials, httpClient: _httpClient);
+      _sqs ??= Sqs(credentials: _credentials, httpClient: _httpClient);
 
   /// Returns an SNS service, inheriting the properties of this instance.
-  Sns sns(String region) => new Sns(
+  Sns sns(String region) => Sns(
       credentials: _credentials, httpClient: _httpClient, region: region);
 
   Lambda lambda(String region) => Lambda(
       credentials: _credentials, httpClient: _httpClient, region: region);
+
+  DynamoDB dynamoDB(String region) => DynamoDB(region: region, credentials: _credentials);
 }
--- a/lib/dynamodb/dynamodb-2012-08-10.dart
+++ b/lib/dynamodb/dynamodb-2012-08-10.dart
@@ -42,34 +42,33 @@ class DynamoDB {
   Credentials credentials;
   final Map<String, String> environment = Platform.environment;
 
-  DynamoDB({this.region = 'eu-west-1', this.credentials}) {
-    dio = Dio();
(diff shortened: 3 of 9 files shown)


isoos:
@Schwusch: thanks for writing this, it is really promising. May I ask a few modifications to be done?

1. As the rest of the library uses `package:http_client` (as an instance passed in to `Aws`), please use that instead of `package:dio`.

If needed, I could also add a `dio` proxy to `http_client` (or a similarly named package).

As far as I have checked, the interceptor pattern could be implemented similarly with a class that extends the `Request` object, and that could be shared between these generated APIs. I can give you further help if needed in this.

2. Instead of `Input` and `Output`, I think `Request` and `Response` is better name suffix for parameters.

3. In some cases, the `Input`/`Request` object takes only a few parameters. It would be worth to consider inlining such methods so that the user of the library doesn't need to do too long request object initialization. What do you think?

Schwusch:
When I made the generator, I strived to generate services that were standalone. In my use-case, I often only use one service, or perhaps two.

The Dio package seemed fitting, with less code generation from my part and pleasant experiences from previous projects. I recommend the package to everyone. The signing function could be imported and shared though.

If you add a proxy, I could take Dio as an optional parameter. I would not recommend sharing clients between the services though.

I generate the services from this list of definitions: https://github.com/aws/aws-sdk-js/tree/master/apis

The naming convention is defined in those JSON files, and corresponds to the documentation. I've already omitted names for simple types,  such as [`AttributeAction`](https://github.com/aws/aws-sdk-js/blob/master/apis/dynamodb-2012-08-10.normal.json#L1030) which is a `String`. Dart does not allow `typedef`s  for anything but functions anyway.

I could do a "find and replace" for `Input` -> `Request` and `Output` -> `Response`. That would correspond better to [the Java SDK](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/AbstractAmazonDynamoDB.java#L42). The naming convention [differs](https://github.com/aws/aws-sdk-js/blob/master/apis/lambda-2015-03-31.normal.json#L2266) between the services and the calls within the service though.

In the generator, it is hard to draw a line where the amount of members in the input structure could be split up into separate arguments. Especially when some goes in the headers, some in the path, and some in the request body. This is most true in the `lambda` api, as I've seen so far.

isoos:
I'll need to check `package:dio` in-depth how it handles some of the nuances of HTTP clients (should be done by end-of-day today).

Until then, would you consider sharing the code generator too?

Schwusch:
I have no prior experience in code generation, so it is kind of messy. 
But here it is: https://github.com/Schwusch/aws_sdk_dart

isoos:
> I would not recommend sharing clients between the services though.

Why is that?

The current design passes the http client instance around from the top-level `Aws` instance to the other, that way if you have initialized you custom http client, it will be applied across all the services. I would like to not break that behaviour (e.g. it allows easy initialization of using a http proxy for everything, or to have a client that keeps the http connections for a while, but after N request it creates new ones).

Schwusch:
> Why is that?

I've hit bugs when sharing http clients before. It's mostly headers, tokens/cookies and cache-settings that is auto-stored in the client, that is not applicable to other services and will introduce bugs. If there is a way to isolate the subdomains from affecting each other, then I see no problem. 
Each service just have to be more explicit in defining the url, since the base url cannot be shared.

Schwusch:
I've come to realise the code generation approach might be unreasonable complex. When studying the JS SDK, I see that their approach is mitigating complexity by offloading most logic to [helper methods](https://github.com/aws/aws-sdk-js/blob/master/lib/core.js) operating on the [JSON APIs](https://github.com/aws/aws-sdk-js/blob/master/lib/api_loader.js). Perhaps such a solution is preferable, and static typing could be a last step, generated by the method I've used so far.

What do you think?

isoos:
> It's mostly headers, tokens/cookies and cache-settings that is auto-stored in the client, that is not applicable to other services and will introduce bugs.

The benefit of `package:http_client` is that it does not store random settings, session state, not even the base URL. It is the minimal HTTP client that libraries like AWS client should probably use. I'm not convinced that the user-friendly part of `dio` is required here.

> When studying the JS SDK, I see that their approach is mitigating complexity by offloading most logic to helper methods operating on the JSON APIs, like this example. Perhaps such a solution is preferable, and static typing could be a last step, generated by the method I've used so far.

I mostly agree, although a generated API is usually better than no API. Maybe we could borrow some concepts from `package:gcloud` and its use of `package:googleapis`. The later is a generated package (using `package:http` as the transport library), and the former is a high-level, developer-friendly Dart-based API on top of that.

Schwusch:
Perhaps it is better discussed as an issue, rather than a PR. I'm not sure how to approach the subject though.

Schwusch:
I started a discussion in an issue instead: https://github.com/agilord/aws_client/issues/15

PR #16, opened 2019-11-20
Draft of code generation

Schwusch:
This generates compiling, but non functional code.
The generated code is a shell, in which we will put generic business logic in the future.

In order to generate, open a terminal in the `bin` directory, and run:
```bash
dart generate.dart download
bash generateFromDownloaded.sh
```

This will download and transform JSON files from the aws-sdk-js repository, generate JSON serialization and format the code nicely.


Schwusch:
> Take a look at `CommandRunner`
Thanks, I will do exactly that. The `run build_runner build` and `dartfmt` comands will fit nicely in there.
It now has a subbcommand, `dart generate.dart generate -d` for downloading before generating, it will tell you in the `--help` documentation.

I will update the `README`

Schwusch:
I misunderstood the CommandRunner. I thought it did what `Process.run()` is doing.

I'm not sure what value it brings over the current solution. Perhaps if the generator grows in size and functionality it helps organizing the flow. But right now it only does two things: download and/or generate.

PR #17, opened 2019-11-20
Split generator directories and project files.

isoos:
- Moved package files to `aws_client/` (updated dependencies in `pubspec.yaml`)
- Moved generator files to `generator/` (added `pubspec.yaml`)
- Added top-level `README.md`


Schwusch:
LGTM

Schwusch:
Have you verified, running the generator after the split?

isoos:
> Have you verified, running the generator after the split?

Not yet. I'll do the CommandRunner update too, and will do the verification with that.

Schwusch:
I suspect this line: https://github.com/agilord/aws_client/blob/d84dc772a90d903f3e0eed262efd464c8c19b8e1/generator/bin/library_builder.dart#L36
will have to be changed.

isoos:
Ah, yeah, I guess that should be relative to the project directory. I'll update that separately.

PR #18, opened 2019-11-21
Apply CommandRunner

isoos:


isoos:
/cc @Schwusch

The diff view doesn't make much justice to it, but you can probably see that a lot of the parsing and if-then handling can be loaded off to the `CommandRunner` pattern.

isoos:
(If you do turn off displaying the whitespace changes, the diff becomes much better.)

Schwusch:
Yupp, that makes it very readable. I copied the `package:googleapis` style, which doesn't have to be best practice.

PR #19, opened 2019-11-21
Refactor commands.

isoos:
- moving things outside of `bin/` (eventually becomes more testable)
- fixed path


Schwusch:
Looks good!

PR #20, opened 2019-11-21
Fix paths.

isoos:


isoos:
@Schwusch: I've now fixed the paths and verified that everything is working.

PR #21, opened 2019-11-24
Add protocol stubs, make JSON specification const

Schwusch:
I've added stubs of all the service protocols declared in `aws-sdk-js`. I've also added the protocol calls in `library_builder.dart`.

### DIFF
--- a/aws_client/lib/src/json/builder.dart
+++ b/aws_client/lib/src/json/builder.dart
@@ -1,0 +1,29 @@
+import 'dart:convert';
+
+class JsonBuilder {
+  String build(Map<String, dynamic> value, Map<String, dynamic> shape) =>
+      jsonEncode(_translate(value, shape));
+
+  dynamic _translate(Map<String, dynamic> value, Map<String, dynamic> shape) {
+    if (shape == null || value == null) return null;
+
+    switch (shape['type'] as String) {
+      case 'structure':
+        return _translateStructure();
+      case 'map':
+        return _translateMap();
+      case 'list':
+        return _translateList();
+      default:
+        return _translateScalar();
+    }
+  }
+
+  Map<String, dynamic> _translateStructure() => null;
+
+  Map<dynamic, dynamic> _translateMap() => null;
+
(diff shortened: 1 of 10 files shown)


aws_client/lib/src/json/builder.dart
```diff
@@ -0,0 +1,29 @@
+import 'dart:convert';
+
+class JsonBuilder {
+  String build(Map<String, dynamic> value, Map<String, dynamic> shape) =>
+      jsonEncode(_translate(value, shape));
+
+  dynamic _translate(Map<String, dynamic> value, Map<String, dynamic> shape) {
```
isoos: I'm not sure I understand the need for this. Could you please link into the JS implementation what this supposed to be doing?

generator/lib/library_builder.dart
```diff
@@ -206,10 +205,8 @@ class $className {""");
 
     writeln("""
     buildRequest(
-      RequestWrapper(
         spec,
-        ${parameterMembers.isNotEmpty ? "param.toJson()..addAll({'operation': '$methodName'})," : "{'operation': '$methodName'}"}
-      ),
+        ${parameterMembers.isNotEmpty ? "param.toJson()..addAll({'operation': '$methodName'})" : "{'operation': '$methodName'}"},
```
isoos: Maybe simpler:
```dart
{
  ...param.toJson(),
  'operation': methodName,
}
```

isoos:
Merging this for now, let's iterate on it as we go...

generator/lib/library_builder.dart
Schwusch: I did not know that was possible, nice catch

aws_client/lib/src/json/builder.dart
Schwusch: Sure, https://github.com/aws/aws-sdk-js/blob/eaf5f577b7b719c055aa72f6e40cf606fb72c34e/lib/json/builder.js#L3

PR #25, opened 2019-11-29
Simpler member names.

isoos:
At first look there were no member names that were colliding with other parts of these classes or Dart reserved words. Let's remove the `m` prefix from their name.

@Schwusch: wdyt?

Oh, and I've also updated the timestamp's Dart type, as `json_serializable` should handle it, assuming it is the standard ISO 8601 format. (is it?)

### DIFF
--- a/generator/lib/library_builder.dart
+++ b/generator/lib/library_builder.dart
@@ -145,7 +145,7 @@ extension StringStuff on String {
       case 'blob':
         return 'blob';
       case 'timestamp':
-        return 'String';
+        return 'DateTime';
       default:
         return '???';
     }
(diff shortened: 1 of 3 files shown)


isoos:
Oh, I've noticed that there are a few members that are reserved words, e.g. `return` and `default`. Added logic to use `returnValue` and `defaultValue` in such cases.

Schwusch:
I've fought a lot with member names. As you mentioned, some become reserved words, some have names same as the type and some are capitals only. The lazy solution was to put `m` in front.

I am unsure of which format the timestamp is. Time will tell hopefully

generator/lib/library_builder.dart
```diff
@@ -171,17 +171,16 @@ class $className {''');
...
+        returnShape.hasEmptyMembers) {
       returnType = 'void';
     }
     final input = method.input;
     final parameterType = input?.shape;
 
-    final Map<String, Member> parameterMembers =
-        (shapes[parameterType]?.members ?? <String, Member>{});
+    final parameterShape = shapes[parameterType];
+    final useParameter = parameterShape != null && parameterShape.hasMembers;
```
Schwusch: I usually do `parametershape?.hasMembers ?? false;` but it's a matter of taste

isoos:
rebased

generator/lib/library_builder.dart
isoos: Good point. I'll do it in a follow-up.

PR #27, opened 2019-12-01
First if-else for protocol switch.

isoos:
The PR also makes the method names lowercase.


isoos:
@Schwusch: I'll have some time in the evening to implement the query protocol. We already have the `sns` library implemented by hand, it should be a good start for it.

Schwusch:
That looks great. I was about to suggest that we implement those protocols already made in the handwritten library. If you make a reference implementation for one protocol, I can follow your style with another.

I look forward to it, it would be a milestone to have a working service!

PR #28, opened 2019-12-01
First steps towards a query protocol impl.

isoos:
Code sample:

````dart
class SNS {
  final QueryProtocol _protocol;

  SNS({Client client}) : _protocol = QueryProtocol(client: client);

  Future<void> addPermission(AddPermissionInput input) async {
    ArgumentError.checkNotNull(input, 'input');
    final $request = <String, String>{
      'Action': 'AddPermission',
      'Version': '2010-03-31',
      ...input.toJson(),
    };
    await _protocol.send($request, method: 'POST', requestUri: '/');
  }
````

There are a few gaps with TODOs, and the XML processing may not be the 100% correct, but let's iterate on this one.

### DIFF
--- a/generator/lib/library_builder.dart
+++ b/generator/lib/library_builder.dart
@@ -28,7 +28,7 @@ import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
 
 """);
-  if (api.metadata.protocol == 'query') {
+  if (api.usesQueryProtocol) {
     buf.writeln('import \'package:aws_client/src/protocol/query.dart\';');
   }
   buf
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: generator/lib/library_builder.dart [hunk 1]
```diff
@@ -158,7 +160,11 @@ extension StringBufferStuff on StringBuffer {
 
     writeln('''
 ${api.documentation.splitToComment()}
-class $className {''');
+class $className {
+  final QueryProtocol _protocol;
+
+  $className({Client client}) : _protocol = QueryProtocol(client: client);
```
Schwusch: Shouldn't this be conditional like above, `if (api.metadata.protocol == 'query') `  ?

generator/lib/library_builder.dart
```diff
@@ -189,14 +196,39 @@ class $className {''');
...
 
-    writeln(
-        "  Future<$returnType> ${operation.methodName}(${useParameter ? "$parameterType param" : ""}) async {");
+    write('  Future<$returnType> ${operation.methodName}(');
+    if (useParameter) write('$parameterType input');
+//    TODO: migrate to per-member input parameters
+//    if (useParameter) write('{');
+//    for (final member in parameterShape?.members ?? <Member>[]) {
+//    }
+//    if (useParameter) write('}');
```
Schwusch: I like this idea, let's see how it plays out

generator/lib/library_builder.dart
isoos: good catch, fixing!

PR #30, opened 2019-12-03
Misc fixes: returnType, fieldName

isoos:
- Fixes where the `Future<void>` method still tried to return with a `.fromJson()` (single `returnType`)
- Fixes weird names, e.g. `aWSAccessToken` -> `awsAccessToken`.
- I've added reference to the top-level `Api` in `Shape`, `Operation` and `Descriptor`, as I'd like to unify and move the logic that generates the dart types (e.g. `List<Whatever>`).

### DIFF
--- a/generator/lib/model/operation.dart
+++ b/generator/lib/model/operation.dart
@@ -53,8 +53,8 @@ class Operation {
       _$OperationFromJson(json);
 
   void initReferences() {
-    input.api = api;
-    output.api = api;
+    input?.api = api;
+    output?.api = api;
     errors?.forEach((e) => e.api = api);
   }
 
--- a/generator/lib/model/operation.dart
+++ b/generator/lib/model/operation.dart
@@ -1,11 +1,14 @@
 import 'package:json_annotation/json_annotation.dart';
 
+import 'api.dart';
 import 'descriptor.dart';
 
 part 'operation.g.dart';
 
 @JsonSerializable(createToJson: false, disallowUnrecognizedKeys: true)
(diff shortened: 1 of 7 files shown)


isoos:
I'm sorry, the `dartType`-centric refactor did creep into this PR a bit too much, but I think it is an important change to do, although mostly unrelated to the other fixes.

generator/lib/model/operation.dart
```diff
@@ -1,11 +1,14 @@
...
 
+import 'api.dart';
 import 'descriptor.dart';
 
 part 'operation.g.dart';
 
 @JsonSerializable(createToJson: false, disallowUnrecognizedKeys: true)
 class Operation {
+  @JsonKey(ignore: true)
+  Api api;
```
Schwusch: The garbage collector is probably not happy with this one, but it is not a long running program anyway 

¯\\\_(ツ)\_/¯

isoos: I don't see a case where `Operation` is GC-d, but its containing `Api` is not :)

btw. the Dart VM is really good in memory allocation, and also in GC (the single-threaded isolate + strongly typed memory model has its benefits).

Schwusch: Yeah, I read up on mark & sweep, it should be fine, we let go of the `Api` instance in [`generate.dart`](https://github.com/agilord/aws_client/blob/master/generator/lib/generate_command.dart#L58). It should handle circular references since it is not attached to a GC root.

PR #31, opened 2019-12-04
Adds options to not run build_runner and dartfmt.

Schwusch:
It is nice to skip these in the dev cycle, to iterate faster.

### DIFF
--- a/generator/lib/generate_command.dart
+++ b/generator/lib/generate_command.dart
@@ -22,6 +22,20 @@ class GenerateCommand extends Command {
         'download',
         abbr: 'd',
         help: 'Downloads the definitions first before generating',
+      )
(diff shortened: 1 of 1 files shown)


generator/lib/generate_command.dart
```diff
@@ -22,6 +22,20 @@ class GenerateCommand extends Command {
         'download',
         abbr: 'd',
         help: 'Downloads the definitions first before generating',
+      )
+      ..addFlag(
+        'build_runner',
```
isoos: I think `build-runner` would be a better name for it, as the command line would be `--no-build-runner` vs. `--no-build_runner`.

Schwusch: that's true

PR #32, opened 2019-12-04
Migrate to per-member input parameters, don't generate unused classes

Schwusch:


generator/lib/builders/query_builder.dart
```diff
@@ -21,16 +21,19 @@ class QueryServiceBuilder extends ServiceBuilder {
...
         '      \'Version\': \'${api.metadata.apiVersion}\',');
-    if (useParameter) {
-      buf.writeln('      ...input.toJson(),');
-    }
     buf.writeln('    };');
+    parameterShape?.members?.forEach((member) {
+      if (member.isRequired) {
+        buf.writeln("\$request['${member.name}'] = ${member.fieldName};");
+      } else {
+        buf.writeln("${member.fieldName}?.also((arg) => \$request['${member.name}'] = arg);");
```
isoos: I'd prefer an explicit `if (${member.fieldName} != null) { ...}` block instead. It is a bit more boring, but it feels more readable to me.

Schwusch: It is my Kotlin nerve twitching. I can change it to good ol' if statements. Don't hesitate to change something yourself if you feel like it, I have turned on permission for you to add commits.

isoos: In the end I can live with that, it can be updated later too :)

isoos:
Please fix the CI issue, I believe it is a linter error..

PR #34, opened 2019-12-05
Throw Exception on error response

isoos:


Schwusch:
Should we use the `errors` available in the metadata somehow at some point? I don't really know how to leverage them or what they mean though.

```
"operations": {
    "AddPermission": {
      "name": "AddPermission",
      "http": {
        "method": "POST",
        "requestUri": "/"
      },
      "input": {
        "shape": "AddPermissionInput"
      },
      "errors": [
        {
          "shape": "InvalidParameterException"
        },
        {
          "shape": "InternalErrorException"
        },
        {
          "shape": "AuthorizationErrorException"
        },
        {
          "shape": "NotFoundException"
        }
      ],
      "documentation": "<p>Adds a statement to a topic's access control policy, granting access for the specified AWS accounts to the specified actions.</p>"
    },
```

isoos:
> Should we use the errors available in the metadata somehow at some point?

That's a good point, maybe we should generate Exception classes that are subclass of `AwsException`. The actual throwing part is a bit tricky, but should be doable....

isoos:
Looks like `package:pedantic` got upgraded, I'll update the code in this PR.

PR #35, opened 2019-12-05
Typed exceptions.

isoos:


generator/lib/library_builder.dart
```diff
@@ -81,6 +84,10 @@ ${builder.constructor()}
       writeln("@Deprecated('Deprecated')");
     }
 
+    operation?.errors?.map((d) => d.shape)?.forEach((e) {
+      writeln('  /// May throw $e.');
```
Schwusch: This would be more documenting, like
```
writeln('  /// May throw [$e].');
```
Since it points to an actual class

isoos:
> Perhaps we should start to generate tests as well?

I wouldn't go that far yet, first, let's make this code working :)

PR #36, opened 2019-12-05
Adding scaffolds for JSON based protocols.

Schwusch:
Also, replacing JS blob type with Uint8List, removing the need for Base64Converter


isoos:
There are unused variables that are lint errors. do a print('$var'); to silence that...

Schwusch:
oh, I misunderstood, fixing

PR #38, opened 2019-12-07
Fix build_runner errors

Schwusch:
Put Uint8ListConverter at intended member, add converter for List\<Uint8List\>

### DIFF
--- a/generator/lib/library_builder.dart
+++ b/generator/lib/library_builder.dart
@@ -34,6 +34,8 @@ File buildService(Api api) {
 
   final buf = StringBuffer()..writeln("""
 // ignore_for_file: unused_import
+// ignore_for_file: unused_local_variable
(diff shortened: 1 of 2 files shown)


generator/lib/library_builder.dart
```diff
@@ -142,6 +142,12 @@ ${builder.constructor()}
             writeln("/// Possible values: [${valueEnum.join(", ")}]");
           }
 
+          if (member.dartType == 'Uint8List') {
+            writeln('@Uint8ListConverter()');
+          } else if (member.dartType == 'List<Uint8List>') {
```
isoos: Oh, we have such these on query?

Schwusch: I've only seen them on DynamoDB so far.

PR #39, opened 2019-12-07
Flatten query parameters.

isoos:
I may misinterpret [query_param_serializer.js ](https://github.com/aws/aws-sdk-js/blob/master/lib/query/query_param_serializer.js), @Schwusch could you please take a second look if this looks about right?


Schwusch:
I'm thinking about writing some tests and comparing the results with the JS version ¯\\\_(ツ)\_\/¯
It's hard to know beforehand how accurate it is

isoos:
Fair point on the tests, I've started adding a few, and already found bugs :)
On the other hand, the JS SDK is not strong on tests: https://github.com/aws/aws-sdk-js/blob/master/test/protocol/query.spec.js#L62-L115

PR #40, opened 2019-12-07
Sign query-protocol messages.

isoos:
@Schwusch: tested with SQS, `receiveMessage` request works, however, the result processing is still broken.


isoos:
I've updated the XML -> Map conversion code a bit, but it does not work with flattened list structures. I'm merging the PR as it is, because it is still an improvement with the signing.

PR #42, opened 2019-12-14
Generage [shape].fromXml() constructors for query protocol.

isoos:
There may be edge cases left, but I've verified this with SQS's receive message operation, and it was working as expected.


isoos:
[I'll merge this so you can try it out, but will update the code on review comments.]

PR #45, opened 2019-12-15
Isolate XML marshalling to XML based services

Schwusch:
Other services protocols were broken, e.g. DynamoDB.
Fixed by only generating `fromXml()` in XML services.


isoos:
I'd prefer #46 over this one (it was already in the works), because two protocols require it.

Schwusch:
Yeah, #46 looks good. This can be closed

PR #46, opened 2019-12-15
Restrict fromXml use to query and rest-xml protocols.

isoos:


generator/lib/library_builder.dart
```diff
@@ -45,7 +45,7 @@ import 'dart:typed_data';
 import 'package:http/http.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
-import 'package:xml/xml.dart';
+${api.usesXml ? 'import \'package:xml/xml.dart\';' : ''}
```
Schwusch: This could probably be moved to `QueryServiceBuilder.imports()` to minimize clutter in the main builder

PR #47, opened 2019-12-15
Move package:xml import to query_builder

isoos:


isoos:
@Schwusch like this?

Schwusch:
That's what I had in mind, yes :)

PR #48, opened 2019-12-15
Fallback file name when metadata.uid is missing.

isoos:
- `lambda-2014-11-11` is one of the example of this
- I've added a TODO to lowercase both the directory and the filename, because IIRC it may cause issues on Windows. The exact detail (e.g. `applicationdirectoryservice` vs `application-directory-service` vs `ads`) is up to decide later, these are just placeholders now.


Schwusch:
That works just fine

PR #50, opened 2019-12-15
Better enum value generation.

isoos:
- abstract class
- lowercase field names, only contains allowed characters
- `$` prefix for reserved words (e.g. `$bool`) and numbers (e.g. `$3_6`).
- empty shape members are rendered too (some API methods use them)

### DIFF
--- a/generator/lib/model/shape.dart
+++ b/generator/lib/model/shape.dart
@@ -194,11 +194,21 @@ class Member {
   }
 }
 
-extension on String {
+extension NameStuff on String {
   bool get isAllUpperCase => toUpperCase() == this;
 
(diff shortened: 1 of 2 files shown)


generator/lib/model/shape.dart
```diff
@@ -194,11 +194,21 @@ class Member {
   }
 }
 
-extension on String {
+extension NameStuff on String {
```
Schwusch: FYI, this makes the extension not-private

isoos: Yes, I'm using it elsewhere. I'm planning to consolidate these extensions, as it is getting hard to track which string method exists in which scope.

PR #51, opened 2019-12-15
Fix generated exception classes (most of them).

isoos:
This is only a temporary fix, but it handles 98% of the exception errors we have with the current state. I think it would be better to have only a very limited `AwsException` class in the long run, and the generated classes expose their content as generated. But that is for a follow-up PR, this fix is badly needed.

### DIFF
--- a/generator/lib/library_builder.dart
+++ b/generator/lib/library_builder.dart
@@ -145,8 +145,14 @@ ${builder.constructor()}
       }
       writeln('@JsonSerializable(includeIfNull: false, explicitToJson: true)');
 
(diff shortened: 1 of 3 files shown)


Schwusch:
I assume this is to remove duplicates? Or what does this change do?

isoos:
> I assume this is to remove duplicates? Or what does this change do?

Removes duplicates + also extends shapes that are part of the protocol messages (which may have message and/or code attribute too).

Schwusch:
I see. I guess that is why there is error-extracting functions in JS. 
Nice work

PR #52, opened 2019-12-15
Draft of rest-json protocol

Schwusch:
I'm looking at Lambda to begin with, following the JS SDK as much as possible. 
Not finished by any means yet

### DIFF
--- a/aws_client/lib/src/protocol/rest-json.dart
+++ b/aws_client/lib/src/protocol/rest-json.dart
@@ -1,13 +1,61 @@
 import 'package:http/http.dart';
+import 'package:meta/meta.dart';
+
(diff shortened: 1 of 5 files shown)


isoos:
Can you please do `dartfmt` too?

Schwusch:
dartfmt didn't complain ¯\\\_(ツ)\_/¯
Is there any way to bake in dartanalyzer into the dartfmt command?

isoos:
> Is there any way to bake in dartanalyzer into the dartfmt command?

Not that I know of. I'm using IntelliJ IDEA Community Edition, and it always displays the current analyzer issues, so I don't need to run it separately.

PR #56, opened 2019-12-18
Fix duplicate reference of Credentials via library prefix

isoos:
There were only two errors, but still...


isoos:
`sts-2011-06-15.dart` and `connect-2017-08-08.dart`

PR #57, opened 2019-12-18
Prevent name collision of 'payload' member in rest_json_builder.

isoos:
I think there were two errors overall.


isoos:
in `lex-models-2017-04-19.dart`

isoos:
> Your dartanalyzer must be running white-hot. Nice catch

Just doing a low-effort cleanup today :)

PR #60, opened 2019-12-18
Fix: Shape.isNotUsed

isoos:
The previous heuristic was not correct in all cases, instead, let's use a top-to-bottom mark of what is used.


isoos:
(the last one for today)

PR #66, opened 2019-12-21
Half of the rest-xml skeleton.

isoos:
This is the non-controversial part of the rest-xml protocol skeleton. I'm planning to reduce the code duplication after merging this.

The next part needs to decide how we want to (a) create the request and also (b) parse the response, as e.g. `fromXml` is not enough, as result headers need to be added too.

### DIFF
--- a/aws_client/lib/src/protocol/rest-xml.dart
+++ b/aws_client/lib/src/protocol/rest-xml.dart
@@ -1,5 +1,61 @@
-import 'package:http_client/http_client.dart';
+import 'package:http/http.dart';
+import 'package:meta/meta.dart';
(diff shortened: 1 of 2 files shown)


Schwusch:
The code looks good. Regarding the problem with handling input and output parameters/data, it seems to exist at three places. The URLs, headers and body. Constructing a request would need those three parts from the generated code. Generated parsers would need to accept both body and headers for completeness. Am I thinking about this in the right way?

I made partial implementations of the URL and header construction in rest-json, which can mostly be seen in `Lambda`.

isoos:
Yeah, you're right, we also need the URL too.

PR #67, opened 2019-12-21
Shared header and uri calculation.

isoos:
also updated:
- handling `null` header values
- handling `/{Key+}` URL pattern
- fixed `Map<String, String>` in headers
- fixed header key name


Schwusch:
Nice, LGTM

PR #69, opened 2019-12-22
Process result headers in rest-xml

isoos:
I'm not sure if this approach maps into rest-json as well, but it might if we would do the codegen for the JSON encode and decode too.


Schwusch:
Impressive :) We'll iterate if it doesn't suit json/json-rest

PR #75, opened 2019-12-28
Generate packages

Schwusch:
The `build_runner` step isn't feasible anymore, since the services are separate packages/projects.
Before running `pub run build_runner build` in a project, `pub get` has to be run first. 
This could of course be done for each project, but I think that is another step, e.g. when inspecting that particular package or when publishing.

There seems to be no way to dump JSON as YAML in Dart, so we will have to do with JSON for now.
I've verified that the generated packages work/gives no errors after `pub get && pub run build_runner build` has been run in the package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,4 @@ doc/api/
 *.iws
 
 # Generated directories
-**/generated/
 **/apis/
--- a/generator/config.yaml
+++ b/generator/config.yaml
@@ -1,0 +1,15 @@
+# Shared library version for each protocol
+shared_version:
+  json: 0.3.0-dev
+  rest-json: 0.3.0-dev
+  rest-xml: 0.3.0-dev
+  query: 0.3.0-dev
+  ec2: 0.3.0-dev
+
+# Packages to generate
+packages:
+  - aws_lambda_api # rest-json
+  - aws_dynamodb_api # json
+  - aws_sns_api # query
+  - aws_route53_api # rest-xml
+  - aws_ec2_api # ec2
--- a/generator/lib/builders/ec2_builder.dart
+++ b/generator/lib/builders/ec2_builder.dart
@@ -11,10 +11,10 @@ class Ec2ServiceBuilder extends ServiceBuilder {
   String constructor() => '';
 
   @override
-  String imports() => '';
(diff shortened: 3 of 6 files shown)


isoos:
Let's not do the "Put generated packages under version control" yet. I'd like to put them under the generated package name which I merged in the other PR, but before doing that, I think we should explore a few questions (part of which you have addressed here):

The `pub get` part could be handled via [package:mono_repo](https://pub.dev/packages/mono_repo), and it also generates a single big travis file that we can use. However, we still need to figure out how to correctly increment versions and track changes in `generated_package/CHANGELOG.md`, e.g. if we change something for `protocol:query`, it should only trigger an update in those packages, and the rest can remain the same.

I like your approach of bumping package version if it differs, it would be nice if it could be combined with the corresponding entry in `CHANGELOG.md` too.

However, my suggestion is to take this PR apart and let's do it in smaller steps:
- let's pick only a few packages that we put under version control, ideally one for each protocol, and using the lowercased package name as directory (so that we don't need to rename/move them later).
- let's figure out the per-package boilerplates (`version` in `pubspec.yaml` and in `CHANGELOG.md`, travis, track what and when to publish)
- I have no strong opinion on `dartfmt` vs `package:dart_style`, you could submit that change independently of the above.
- let's figure out some tests that could be automatically added to these package sources (e.g. typically they should not contain the string literal `'null'`, as it usually happens when a `null` reference is passed in a string expression), and add a few of these to the packages
- once we are happy with the above, let's start publishing the ones which we think are in a reasonably acceptable state, and also add the other generated packages for the given protocol

wdyt about the above?

isoos:
Another thing: I think the shared library should be referenced in the pubspec as a hosted (non-local) reference, and only using local path in the `dependency_overrides`. Maybe it will be a config option in the code generator to switch between the two.

Schwusch:
What you say makes sense.
I've remade the PR, squashed the changes and added some of your suggestions.
I'm not sure what your intentions with the tests are exactly, so I've left them out.
I'm not following the `CHANGELOG.md` reasoning, so I haven't done that one. I'm guessing you mean putting commit messages in the changelog? I will let you define that further.

Let's continue this discussion in https://github.com/agilord/aws_client/pull/76

PR #76, opened 2019-12-29
Generator reads config file

Schwusch:
- The generator now reads a config file, much like `package:googleapis`.
- The config file specifies which services to generate
- The config file specifies which version of the shared package each protocol should use, uses local path in `dependency_overrides` when in `--dev`-mode.
- Uses dart_style programmatically instead of separate `dartfmt`.
- Bumps version in package when generated source differ from old source,
or when shared version in config differs and `--bump` is given as argument
- Generates packages

### DIFF
--- a/generator/lib/generate_command.dart
+++ b/generator/lib/generate_command.dart
@@ -3,7 +3,10 @@ import 'dart:io';
 
 import 'package:args/command_runner.dart';
 import 'package:aws_client.generator/model/api.dart';
+import 'package:dart_style/dart_style.dart';
 import 'package:json_annotation/json_annotation.dart';
+import 'package:version/version.dart';
+import 'package:yaml/yaml.dart';
 
 import 'download_command.dart';
 import 'library_builder.dart';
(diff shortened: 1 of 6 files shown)


generator/lib/generate_command.dart
```diff
@@ -23,100 +26,150 @@ class GenerateCommand extends Command {
...
         'format',
         abbr: 'f',
         help: 'Runs dartfmt on the generated code',
         defaultsTo: true,
         negatable: true,
+      )
+      ..addFlag(
+        'bump',
+        abbr: 'b',
+        help: 'Automatically bump package versions on code changes',
```
isoos: I think this is a good start, I have no quick alternative solution.
On long term, I'd like have a `--changelog "<message>"` CLI option. If it is set, we should also add the `"<message>"` under `CHANGELOG.md`, under the new version. The complexity comes from the fact that we may need to add multiple entries under the same version, so it would either need to create the version upfront, or do some complex calculation on which cases we don't increment the version just merge the changelog entry under the current pending one.

generator/lib/generate_command.dart
```diff
@@ -23,100 +26,150 @@ class GenerateCommand extends Command {
...
+      )
+      ..addFlag('dev',
```
isoos: I'm wondering if `dev` should be the default instead of a separate flag, and we should have a linting test that checks that we don't checkin `package.yaml` files with local overrides.

Again, let's start with this version, and we'll see how it goes.

generator/pubspec.yaml
```diff
@@ -19,3 +19,6 @@ dev_dependencies:
   build_verify: ^1.1.1
   json_serializable: ^3.2.0
   test: ^1.0.0
+  version: ^1.0.0
+  dart_style: ^1.3.3
```
isoos: Nit: alphabetical order + newline at the end.

generator/lib/generate_command.dart
Schwusch: Kevin Moore has a wrapper for Git: https://github.com/kevmoo/git

If we write good commit messages and save commit hashes for each version, we could perhaps append all commit messages since last version?

isoos: Hm, true, it would be the perfect solution if it could scan for all the commit messages that affect that package, and put those on the changelog.

PR #77, opened 2019-12-29
Typed config

isoos:
I am not sure how the config file will evolve, but it seems a good idea to make it typed :)


Schwusch:
That is convenient, yes 👍

PR #79, opened 2019-12-29
YAML-formatted pubspec.yaml

isoos:


generator/lib/pubspec_builder.dart
```diff
@@ -0,0 +1,37 @@
...
+  return '''name: ${api.packageName}
+version: $packageVersion
+publish_to: none
+
+environment:
+  sdk: '>=2.6.0 <3.0.0'
+
+dependencies:
+  aws_client: $sharedVersion
+  http: ^0.12.0
```
Schwusch: Just a thought, is `package:http` better to expose as a transitive dependency in the shared package? I'm thinking if we forget to update it here when we update the shared package.

isoos: Good point, however, the generated package library import `package:http`, and with that, we should put it as a direct dependency. As it is mostly a pass-through, we could hide its type with dynamic, in which case the package won't import it, and we can defer it to the shared library. I can see it go both ways, and don't have really strong preference here.

Schwusch: I meant that it could be re-exported in the shared library, so when importing from the shared library the Http client comes with. Or is that a bit unorthodox?

isoos: It is a little bit surprising, but if the `dartanalyzer` does not complain about it, I think it should be fine to do so...

PR #80, opened 2019-12-29
Per-package pub get and build_runner

isoos:


generator/lib/generate_command.dart
```diff
@@ -163,8 +166,43 @@ class GenerateCommand extends Command {
...
       }
-    });
+    }
+
+    for (final baseDir in touchedDirs) {
+      // TODO: once in git, detect if there was no change, and skip when not needed
+      await _runPubGet(baseDir);
+      // TODO: once in git, detect if there was no change, and skip when not needed
+      await _runBuildRunner(baseDir);
+    }
```
Schwusch: Can we have a flag to turn this off? It increases the dev cycle significantly

isoos: Good point, added `--[no-]build`.

PR #84, opened 2019-12-31
Expose build_runner dependencies without prefix

Schwusch:
build_runner doesn't handle import prefixes, show these specifically without prefix

### DIFF
--- a/generator/lib/library_builder.dart
+++ b/generator/lib/library_builder.dart
@@ -162,9 +162,9 @@ ${builder.constructor()}
 
         if (shape.api.generateJson) {
           if (member.dartType == 'Uint8List') {
-            writeln('@shared.Uint8ListConverter()');
+            writeln('@Uint8ListConverter()');
           } else if (member.dartType == 'List<Uint8List>') {
-            writeln('@shared.Uint8ListListConverter()');
+            writeln('@Uint8ListListConverter()');
           }
           writeln("  @JsonKey(name: '${member.name}')");
         }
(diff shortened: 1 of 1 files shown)


Schwusch:
This issue surfaced when I was looking at DynamoDB

generator/lib/library_builder.dart
```diff
@@ -46,6 +46,7 @@ import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart' as _meta;
 
 import 'package:aws_client/shared.dart' as shared;
+import 'package:aws_client/shared.dart' show Uint8ListConverter, Uint8ListListConverter;
```
isoos: Something does not add app: either you'd need to remove the `shared.` prefix from `shared.Uint8ListConverter`, or we are missing the prefix from unprefixed `Uint8ListConverter`. (I'll need to check the head to be sure). What am I missing?

Schwusch: We could remove the `shared.` prefix, but I think you mentioned a clash of naming in some services? Unfortunately `build_runner` does not take prefixes into account when annotating fields like we do:
```dart
@shared.Uint8ListConverter()
@JsonKey(name: 'B')
final Uint8List b;
```
in the `xxx.g.dart` it becomes:
```dart
writeNotNull('B', const Uint8ListConverter().toJson(instance.b));
```

Schwusch: But you're right, it would be more consistent to remove the prefix when annotating

PR #87, opened 2019-12-31
Skeleton code for automatic changelog updates.

isoos:
Ideally it should generate the following log:

```
## [version]

(git hash: [hash])

- Change one #[issue number]
- Change two #[issue number]
```

There are a few things that are not complete, marked with TODOs.


Schwusch:
Nice, it will be really cool with some Git magic :)

PR #88, opened 2019-12-31
Generate formatted documentation for classes and methods.

isoos:
It tries to parse and format the documentation text, which is reasonable for now, but of course it could be further improved later.


isoos:
We'll see how it works :)

PR #91, opened 2020-01-01
Package command

Schwusch:
- read current version
 - increment version per the command line arguments
 - use the correct version in changelog
 - update pubspec.yaml with the new version

### DIFF
--- a/generator/lib/package_command.dart
+++ b/generator/lib/package_command.dart
@@ -2,6 +2,7 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:args/command_runner.dart';
+import 'package:version/version.dart';
 import 'package:yaml/yaml.dart';
 
 import 'model/config.dart';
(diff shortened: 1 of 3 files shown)


generator/lib/package_command.dart
```diff
@@ -53,27 +54,59 @@ class BumpVersionCommand extends Command {
...
+
+      final pubspecFile = File('$pkgDir/pubspec.yaml');
+      final pubspecString = pubspecFile.readAsStringSync();
+      final pubspecMap = json.decode(json.encode(loadYaml(pubspecString)))
+          as Map<String, dynamic>;
+
+      final currentVersion = Version.parse(pubspecMap['version'] as String);
+      Version newVersion;
+
+      if (argResults['version'] == null) {
```
isoos: Note: This was my original intent, but I'm now contemplating how it would be if we encoded this information in the commit messages. E.g. `[major]` or `[minor]` and `[patch]` tags in the commit would trigger the most significant version change, without us needing to specify it explicitly at the time we are running this tool.

Let's go with this for now, but I'd probably consider the commit-driven approach in the long term.

generator/lib/package_command.dart
```diff
@@ -96,16 +128,25 @@ class BumpVersionCommand extends Command {
...
+          );
+      pubspecFile.writeAsStringSync(newPubspecString);
       // TODO: commit new changelog and pubspec.yaml with the updates (+ push?)
     }
   }
 }
 
-Future<List<_Commit>> _listChanges(String dir) async {
-  final pr = await Process.run('git', ['log', '--format=oneline', '--', dir]);
+Future<List<_Commit>> _listChanges() async {
```
isoos: I'm not sure about this part. Do you have any specific need to remove the directory scope here?

I think we should narrow the git changes list to the specific directory. We may have protocols that are unaffected by certain changes which do not require neither a version bump, not a release.

generator/lib/pubspec_builder.dart
```diff
@@ -21,6 +22,8 @@ dependency_overrides:
 version: $packageVersion
 publish_to: none
 
+protocol: $protocol
```
isoos: Nit: Do you expect more of such properties? I'd probably put them under a custom key, e.g.

```
aws:
  protocol: $protocol
```

generator/lib/package_command.dart
Schwusch: Oh, I misunderstood the intent perhaps. The command did not print out anything when creating a new package, and it's not committed.

How do we do this the first time the package is committed?

generator/lib/pubspec_builder.dart
Schwusch: I don't expect more properties. It's the only taxonomy we have, so it made sense to put it there, so we can update the shared package version more effectively.

generator/lib/package_command.dart
isoos: I think the first release can be a generic changelog:
```
## 0.0.1

(git hash: current)

- initial release
```
(or something similar)

isoos: It doesn't even need to have the git hash, just the version and the initial release phrase.

generator/lib/pubspec_builder.dart
Schwusch: I don't think there will be more properties. It's the only package taxonomy we have, and I don't see any others coming.

isoos: If this is the only one we expect here, let's keep it here.

generator/lib/package_command.dart
Schwusch: That is a good idea, and I was thinking about marking the commits with protocol tags depending on which they affect.

generator/lib/pubspec_builder.dart
Schwusch: Funny, GitHub crashed on the first comment so I wrote the second one. And now both shows 🤷‍♂️

PR #94, opened 2020-01-01
Qualified commit messages for Changelog.

isoos:


generator/lib/package_command.dart
```diff
@@ -68,19 +68,36 @@ class BumpVersionCommand extends Command {
...
-            break;
-          default:
-            throw ArgumentError('Unknown "--version-increment" argument');
+        final increment = argResults['version-increment'] as String;
+        final changeMajor =
+            increment == 'major' || currentChanges.any((c) => c.isMajor);
+        final changeMinor =
+            increment == 'minor' || currentChanges.any((c) => c.isMinor);
+        final changePatch =
+            increment == 'patch' || currentChanges.any((c) => c.isPatch);
```
Schwusch: I would also like to add a comparison with the shared package version, like:
```dart
final changeMajor = newSharedVersion.major > oldSharedVersion.major 
// and so on
```

generator/lib/package_command.dart
```diff
@@ -106,22 +117,16 @@ class BumpVersionCommand extends Command {
...
         '## $newVersion',
         if (currentHash?.trim()?.isNotEmpty == true) ...[
           '',
           '(git hash: $currentHash)',
           '',
         ],
-        ...currentChanges.map((c) => '- ${c.message}'),
+        ...currentChanges
+            .where((c) => c.isQualified)
+            .map((c) => '- ${c.formattedMessage}'),
```
Schwusch: In case there is only changes to the shared package, and you want to bump the shared package version, should it be reflected here?

generator/lib/package_command.dart
isoos: I think the changes in the shared library can be independent of the generated API, and it may be even a non-pinned version in the generated library. E.g. depending on shared `^1.0.0` could be a valid shared version dependency for a lot of generated API libraries and their several releases.

On the other hand, we should trigger at minimum a patch increment if it changes. I'll do a follow-up to do this and also update the changelog entries when this happens. (per your other comment)

PR #97, opened 2020-01-02
Move protocol (service)builders.

isoos:
I think `library_builder`, `pubspec_builder` and the new `readme_builder` could go into `builders/` and these should go into their own sub-dir.


Schwusch:
LGTM

PR #102, opened 2020-01-03
Copy LICENSE file to generated projects.

isoos:


Schwusch:
I'm a license illiterate. Is this alright considering the sources we are generating from?  

What are the placeholders intended to be filled with?

isoos:
> Is this alright considering the sources we are generating from?

Good point, I'll need to take a second look on how this works on other projects. It is likely that we need to link to the source files too.

isoos:
Updated PR to copy the AWS SDK license to the generated packages. PTAL.

Schwusch:
LGTM.

PR #106, opened 2020-01-03
Unbundle package:http from shared lib.

isoos:
I think it is better to have it separated from shared, it makes the generated code more readable.

### DIFF
--- a/generator/lib/builders/library_builder.dart
+++ b/generator/lib/builders/library_builder.dart
@@ -44,6 +44,7 @@ String buildService(Api api) {
 import 'dart:convert';
 import 'dart:typed_data';
 
+import 'package:http/http.dart' as _http;
 import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart' as _meta;
 
--- a/generator/lib/builders/protocols/json_builder.dart
+++ b/generator/lib/builders/protocols/json_builder.dart
@@ -10,7 +10,7 @@ class JsonServiceBuilder extends ServiceBuilder {
   @override
   String constructor() => '''
   final shared.JsonProtocol _protocol;
-  ${api.metadata.className}({@_meta.required String region, @_meta.required shared.AwsClientCredentials credentials, shared.Client client, String endpointUrl,})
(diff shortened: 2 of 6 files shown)


Schwusch:
How will the `http` dependency be bumped when `shared_aws_api` upgrades it?

isoos:
> How will the http dependency be bumped when shared_aws_api upgrades it?

e.g. now it is `0.1.0` with `http: ^0.12.0`. When `http: ^0.13.0` happens, we'll bump shared to `0.1.1`. `pub` client will figure out the proper constraints for the users of the library.

isoos:
> How will the http dependency be bumped when shared_aws_api upgrades it?

We could also do `>= 0.12.0 <0.14.0` next time, supporting both (as it is unlikely that the small API surface we are using will have a breaking change).

Schwusch:
Alright but don't forget to add http as a dependency in https://github.com/agilord/aws_client/blob/http-prefix/generator/lib/builders/pubspec_builder.dart

isoos:
> Alright but don't forget to add http as a dependency in https://github.com/agilord/aws_client/blob/http-prefix/generator/lib/builders/pubspec_builder.dart

Ah, fair point. I forgot about that... Hm... maybe this is not the best idea after all...

isoos:
How about copying the dependency constraints from shared?

Schwusch:
Well, it is that kind of mess I thought to mitigate through just letting the http  dependency shine through.

I don't have any ideas how to solve it better. I just want it to be as little maintenance as possible. 

What makes the current solution hard to read?

isoos:
> What makes the current solution hard to read?

I guess it is the consistency what I'd want to improve. Right now we include `package:meta` (as `_meta`), and also `package:json_serializable` (without alias), and they seem to be at a similar level as `package:http`.

Schwusch:
Tell me what you think about https://github.com/agilord/aws_client/pull/107.
I feel like there is no need to explicitly define dependencies twice, since generated packages are so intimately connected to the shared package.

isoos:
#107 seems to be better.

PR #107, opened 2020-01-04
Align transient dependencies, expose through re-exports

Schwusch:
This is an alternative to https://github.com/agilord/aws_client/pull/106

### DIFF
--- a/generator/lib/builders/library_builder.dart
+++ b/generator/lib/builders/library_builder.dart
@@ -44,9 +44,6 @@ String buildService(Api api) {
 import 'dart:convert';
 import 'dart:typed_data';
 
-import 'package:json_annotation/json_annotation.dart';
-import 'package:meta/meta.dart' as _meta;
-
 import 'package:shared_aws_api/shared.dart' as shared;
 import 'package:shared_aws_api/shared.dart' show Uint8ListConverter, Uint8ListListConverter;
 """);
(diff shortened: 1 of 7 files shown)


generator/lib/builders/library_builder.dart
```diff
@@ -44,9 +44,6 @@ String buildService(Api api) {
 import 'dart:convert';
 import 'dart:typed_data';
 
-import 'package:json_annotation/json_annotation.dart';
-import 'package:meta/meta.dart' as _meta;
-
 import 'package:shared_aws_api/shared.dart' as shared;
 import 'package:shared_aws_api/shared.dart' show Uint8ListConverter, Uint8ListListConverter;
```
isoos: I'm wondering if we still need this, but the rest looks good.

Schwusch: What are you referring to? The lines I removed or the ones that are left?

isoos: The unprefixed `show Uint8ListConverter, Uint8ListListConverter`

isoos: The reason being that now json-annotation will be behind the shared prefix too, and maybe that could fix the build_runner issue. Also a side not: we should probably rename it to `_shared` (or `_s`).

PR #109, opened 2020-01-04
Shorter import alias

isoos:


Schwusch:
I'm not overly convinced about the shortening, but I don't think anyone is going to read the source anyway.

I think the underscore is warranted though 😊

PR #110, opened 2020-01-04
Protect against checking in dev mode artifacts.

isoos:


generator/test/no_dev_mode_test.dart
```diff
@@ -0,0 +1,22 @@
...
+        .list(recursive: true)
+        .where((f) => f is File)
+        .cast<File>()
+        .where((f) => basename(f.path) == 'pubspec.yaml')
+        .toList();
+    for (final file in pubspecFiles) {
+      final content = await file.readAsString();
+      expect(content, isNot(contains('dependency_overrides')));
+      expect(content, isNot(contains('../../shared_aws_api')));
+    }
```
Schwusch: Is there any point in converting the iterable to a List, and then iterate on that again? 
I would simply do 
```dart
await Directory('../generated')
        .list(recursive: true)
        .where((f) => f is File)
        .cast<File>()
        .where((f) => basename(f.path) == 'pubspec.yaml')
        .foreach((file) {
          final content = await file.readAsString();
          expect(content, isNot(contains('dependency_overrides')));
          expect(content, isNot(contains('../../shared_aws_api')));
        });

File: generator/test/no_dev_mode_test.dart [hunk 2]
```diff
@@ -0,0 +1,24 @@
...
+    final pubspecFiles = await generatedDir
+        .list(recursive: true)
+        .where((f) => f is File)
+        .cast<File>()
+        .where((f) => basename(f.path) == 'pubspec.yaml')
+        .toList();
+    for (final file in pubspecFiles) {
+      final content = await file.readAsString();
+      expect(content, isNot(contains('dependency_overrides')));
+      expect(content, isNot(contains('../../shared_aws_api')));
```
Schwusch: Is it possible to kill to birds with one stone, and check for `'path:'` instead?

isoos: I'd rather be more specific in what we want to reject.

PR #111, opened 2020-01-04
Publish command

isoos:


generator/lib/package_command.dart
```diff
@@ -188,3 +185,104 @@ class _Commit {
...
+      final protocolConfig = config.protocols[protocol];
+      assert(protocolConfig != null);
+
+      if (!protocolConfig.publish) continue;
+
+      final currentPublishedVersion =
+          await _currentPublishedVersion(client, package);
+
+      if (version == currentPublishedVersion) {
+        // print?
```
Schwusch: I think it would be nice to see a print with e.g. "version x.x.x of aws_foo_api is already published"

generator/lib/package_command.dart
```diff
@@ -188,3 +185,104 @@ class _Commit {
...
+      final protocol = pubspecMap['protocol'] as String;
+      final protocolConfig = config.protocols[protocol];
+      assert(protocolConfig != null);
+
+      if (!protocolConfig.publish) continue;
+
+      final currentPublishedVersion =
+          await _currentPublishedVersion(client, package);
+
+      if (version == currentPublishedVersion) {
```
Schwusch: Question: If one would try to publish a lesser version, lets say 0.1.0 when published version is 0.2.0, what would happen?

isoos: You can, but pub.dev will keep 0.2.0 as the latest stable.

PR #113, opened 2020-01-05
Shallow parameter validation implemented for min, max and pattern.

Schwusch:


shared_aws_api/lib/src/validation/utils.dart
```diff
@@ -0,0 +1,22 @@
+void validateStringPattern(String name, String str, String pattern) {
```
isoos: str -> value

shared_aws_api/lib/src/validation/utils.dart
```diff
@@ -0,0 +1,22 @@
...
+  if (!RegExp(pattern).hasMatch(str)) {
+    throw ArgumentError.value(
+      str,
+      name,
+      'Argument does not conform to $pattern',
+    );
+  }
+}
+
+void validateStringLength(String name, String str, int min, int max) {
```
isoos: str -> value

shared_aws_api/lib/src/validation/utils.dart
isoos: I think all of these need to have an additional null check.

Schwusch:
I suspect that if a parameter is a shape, there can be checks there too. I haven't thought of a good way to check enums yet

shared_aws_api/lib/src/validation/utils.dart
```diff
@@ -0,0 +1,22 @@
...
+  if (min > str.length || str.length > max) {
+    throw ArgumentError.value(
+        str, name, '"$name" is either longer than $max or shorter than $min');
+  }
+}
+
+void validateNumRange(String name, num value, int min, int max) {
```
isoos: Can these be `NaN` / `+inf` or `-inf` by any chance?

Schwusch: `value` can, and I think `NaN` is an outright error so I added tests to verify. `+/-inf` is caught by the existing checks.

shared_aws_api/lib/src/validation/utils.dart
Schwusch: I think `name` being `null` is alright, since it's an optional parameter to `ArgumentError.value()`, the other ones are caught in existing checks

isoos: Oh, sorry, I meant `value` could be null, and we should not throw NPE in the validation code when it happens. (If it is required, the previous `checkNotNull` should handle it).

Schwusch: It is not throwing NPE, I've added [tests for it](https://github.com/agilord/aws_client/blob/ea2b0929f725014b2f69516053b045ce055653bc/shared_aws_api/test/validation/string_validation.dart#L5..L19)

PR #120, opened 2020-01-06
generate example README

Schwusch:


generator/lib/generate_command.dart
```diff
@@ -155,6 +159,7 @@ class GenerateCommand extends Command {
           pubspecFile.writeAsStringSync(pubspecYaml);
           serviceFile.writeAsStringSync(serviceText);
           readmeFile.writeAsStringSync(buildReadmeMd(api));
+          exampleFile.writeAsStringSync(buildExampleReadme(api));
```
isoos: Nit: also need formatting.

Schwusch: Which one, the README or the line in question? I ran `dartfmt` before committing

isoos: Ah, my bad, I thought that it will be a dart file, not a readme.

isoos: I mean *not a markdown...

PR #123, opened 2020-01-06
lowerCaseName member names (esp.enums)

### LINKED ISSUE #122: snake_case conversions
I've noticed that a lot of the enum values have `snake_case` format, which we may transform to `lowerCamelCase` (or whatever the name for it is). For some it works fine, but for e.g. `usEast1` it is debatable if it is better than `us_east_1`). @Schwusch wdyt?

isoos:
I think `lowerCaseNames` are better defaults for long term. We may introduce (explicit) exceptions where we do the non-default choice, but only on a case-by-case basis.

Fixes #122.

### DIFF
--- a/generator/lib/builders/library_builder.dart
+++ b/generator/lib/builders/library_builder.dart
@@ -226,10 +226,6 @@ ${builder.constructor()}
         final shapeClass = member.shapeClass;
         final valueEnum = shapeClass.enumeration;
 
-        if (valueEnum?.isNotEmpty ?? false) {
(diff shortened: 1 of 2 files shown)


Schwusch:
BTW, I do want to point out that enums with numbers separated by underscores become somewhat funny looking.

image

isoos:
I've added another commit to prefix the numbers with `_`, making it into `dotnetcore2_0`. I've also tried using `$`, but `dotnetcore2$0` does not seem to be good for this purpose. wdyt?

Schwusch:
Yeah, that makes it more understandable. Nice!

PR #124, opened 2020-01-06
Enabling query and rest-xml protocols to publish + first check-in.

isoos:


Schwusch:
I'm not shure what `publish: true` does, but the rest looks good

isoos:
> I'm not sure what publish: true does, but the rest looks good

- It removes the `publish_to: none` from `pubspec.yaml` for that protocol.
- It enables publishing through the `package publish` command.

PR #130, opened 2020-01-09
Fix: generate xml-to-enum value for query APIs. [patch]

isoos:
It also makes it easier to generate all the packages for a specific protocol.

### DIFF
--- a/generator/config.yaml
+++ b/generator/config.yaml
@@ -5,7 +5,6 @@ protocols:
     shared: ^0.1.0
   query:
     shared: ^0.1.0
-    generate: true # extend packages list
     publish: true
   rest-json:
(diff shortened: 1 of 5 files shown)


generator/config.yaml
```diff
@@ -5,6 +5,7 @@ protocols:
     shared: ^0.1.0
   query:
     shared: ^0.1.0
+    generate: true # extend packages list
```
Schwusch: Have a look at the print-out from https://github.com/agilord/aws_client/pull/131, when running the generate command.
 It is basically copy/pasting the list to the config.
I'm thinking it's "painting with a very wide brush" when you're not specifying a the services explicitly

Schwusch:
LGTM

PR #133, opened 2020-01-09
Fix rest-xml codegen issues.

isoos:
- the exception should be conditional to the field being non-null, otherwise we don't need to throw an exception (also the extra comma at the end was not good)
- header extraction logic needed a little fix


Schwusch:
How are we handling the generated diff, is it handled in a separate commit after this?

Otherwise, LGTM

isoos:
> How are we handling the generated diff, is it handled in a separate commit after this?

It does not effect `aws_route53_api`, only `aws_s3_api` and `aws_s3control_api`, and they are not in git yet.

PR #134, opened 2020-01-10
Packages with rest-xml protocol.

isoos:


generated/aws_autoscaling_api/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.0.1
+
+(git hash: 9f95ffeaee28d3c62ee0380c32b58f27491840fe)
+
+- initial release
+
```
Schwusch: What is the point with these additions?

generator/config.yaml
```diff
@@ -36,3 +36,7 @@ packages:
   - aws_sns_api
   - aws_sqs_api
   - aws_sts_api
+## rest-xml protocol APIs
+  - aws_cloudfront_api
+  - aws_s3_api
+  - aws_s3control_api
```
Schwusch: Were these not generated thanks to this addition?

generated/aws_autoscaling_api/CHANGELOG.md
isoos: Follows the same pattern that we will be using in the later releases.

isoos:
Ah, wait, something is wrong with this PR, I did something wrong...

PR #137, opened 2020-01-25
Fix: json headers.

isoos:
- Map needs to be `<String, String>`.
- fixed method not returning value on null input
- fixed `dartfmt`


Schwusch:
Have you been able to try the library out, does it work?

isoos:
> Have you been able to try the library out, does it work?

Not yet, these are just analyzer/lint fixes.

PR #145, opened 2020-04-12
Generated Client: MessageAttributeValue and MessageSystemAttributeValue Fix

lucasdina:
I moved the MessageAttributeValue and MessageSystemAttributeValue to the shared project so I could access them from the Query class. Please let me know if theres anything I can/should fix.


isoos:
@lucasdina: I'll take a look on this probably tomorrow, but until then a quick comment:
- `aws_client` is an old codebase, and we are keeping it for reference, but don't actively develop
- `shared_aws_api` is specifically to contain the common code path and interfaces for the AWS protocols. We don't want to move code from generated to shared, as that would not be sustainable in the future.

If there is a bug in the serialization of the generated client, we'll need to fix it by fixing either the shared code or the generator. I haven't checked this PR yet, but based on the moving files from generated to shared, it is unlikely that this will be a correct/future-proof fix.

lucasdina:
@isoos Honestly, I'm still pretty new and have no idea how the generator works. If you point me in the right direction, I'd be more than happy to work on that and submit another PR.

isoos:
@lucasdina: I'm only worried that you invest a lot of time into a solution that has not much chance to get accepted and you lose your good momentum :)

As a start: have you tried to re-generate the sqs library? I think it is a good way to start and tinker (e.g. add a debug statement somewhere in the code generator)...

lucasdina:
@isoos I'll start tinkering around with the project a bit in my free time. For now, I need that SQS code so I'll run my fork the the other project I'm working on and once I can get the generator fixed, I'll switch back to the main repo. Thanks for the pointer!

Issue #144, opened 2020-04-12
[BUG] Error When Flattening MessageAttributeValue and MessageSystemAttributeValue


lucasdina:
When sending an SQS message, the library attempts to flatten the extra parameters from sendMessage() at sqs-2012-11-15.dart:1681 in the auto generated package. The optional parameters are for the messageAttributes (Map<String, MessageAttribute>) and messageSystemAttributes (Map<String, MessageSystemAttribute>). Somewhere down the line, send() gets called at query.dart:42 in the shared auto generated package which calls _buildRequest() > flatQueryParams() > _flatten() all in the same file. _flatten checks for the type to flatten it for the request but fails because there is no case to handle MessageAttribute or MessageSystemAttribute. I would fix and submit another PR but I don't really have the time. Here is the error:

 ```
[VERBOSE-2:ui_dart_state.cc(157)] Unhandled Exception: Invalid argument(s): Unknown type at "MessageAttributes.entry.1.value": MessageAttributeValue (Instance of 'MessageAttributeValue')
#0      _flatten (package:shared_aws_api/src/protocol/query.dart:140:3)
#1      _SyncIterator.moveNext (dart:core-patch/core_patch.dart:146:12)
#2      _SyncIterator.moveNext (dart:core-patch/core_patch.dart:139:32)
#3      _SyncIterator.moveNext (dart:core-patch/core_patch.dart:139:32)
#4      MapMixin.addEntries (dart:collection/maps.dart:172:23)
#5      new Map.fromEntries (dart:core/map.dart:187:17)
#6      flatQueryParams (package:shared_aws_api/src/protocol/query.dart:89:14)
#7      QueryProtocol._buildRequest (package:shared_aws_api/src/protocol/query.dart:74:26)
#8      QueryProtocol.send (package:shared_aws_api/src/protocol/query.dart:49:16)
#9      SQS.sendMessage (package:aws_sqs_api/sqs-2012-11-05.dart:1705:37)
```

EDIT: Here is my code causing the error:
```
    sqs.sendMessage(
        messageBody: unsentEntries[i].toJson().toString(),
        queueUrl: sqsUrl,
        messageGroupId: unsentEntries[i].company.shortName.replaceAll(' ', ''),
        messageDeduplicationId: utils.getUuid(),
        messageAttributes: <String, MessageAttributeValue>{
          'companyId': new MessageAttributeValue(dataType:'string', stringValue: unsentEntries[i].company.id)
        });
```

[...]

PR #146, opened 2020-04-13
Improvements

Schwusch:
Two new services emerge out of this change, Neptune and DocDB. These were placed [under RDS](https://pub.dev/documentation/aws_rds_api/latest/neptune-2014-10-31/Neptune-class.html) before due to sharing the same endpoint prefix. They seem to be vastly different services though.

 - Readme and example is created for the latest version of the APIs.
 - Switch candidate priority in packageBaseName to differentiate
   APIs sharing endpointPrefix.
 - Update Member model class with new fields.

### DIFF
--- a/generator/lib/model/api.dart
+++ b/generator/lib/model/api.dart
@@ -46,16 +46,24 @@ class Api {
   }
 
   bool get usesQueryProtocol => metadata.protocol == 'query';
+
   bool get usesJsonProtocol => metadata.protocol == 'json';
+
   bool get usesRestJsonProtocol => metadata.protocol == 'rest-json';
+
   bool get usesRestXmlProtocol => metadata.protocol == 'rest-xml';
+
   bool get usesEc2Protocol => metadata.protocol == 'ec2';
 
   bool get generateFromJson => usesJsonProtocol || usesRestJsonProtocol;
-  bool get generateToJson => usesJsonProtocol || usesRestJsonProtocol;
+
+  bool get generateToJson =>
+      usesJsonProtocol || usesRestJsonProtocol || usesQueryProtocol;
+
   bool get generateJson => generateFromJson || generateToJson;
 
   bool get generateFromXml => usesQueryProtocol || usesRestXmlProtocol;
+
   bool get generateToXml => usesRestXmlProtocol;
 
   String get directoryPath {
@@ -105,6 +113,7 @@ class Api {
   }
 
   List<String> _exceptions;
+
   List<String> get exceptions {
     if (_exceptions == null) {
       final set = operations?.values
--- a/generator/lib/model/api.dart
+++ b/generator/lib/model/api.dart
@@ -79,9 +79,9 @@ class Api {
 
   String get packageBaseName {
     final candidates = <String>[
-      metadata.endpointPrefix,
       metadata.uid?.split('-20')?.first,
       metadata.className.toLowerCase(),
+      metadata.endpointPrefix
     ];
     final identified = candidates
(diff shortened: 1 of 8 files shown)


Schwusch:
This fixes the crash mentioned in https://github.com/agilord/aws_client/issues/144, but I ran this test:
```dart
import 'package:aws_sqs_api/sqs-2012-11-05.dart';
import 'package:http/http.dart';
import 'package:http/testing.dart';
import 'package:test/test.dart';

void main() {
  test('foo', () async {
    final client = MockClient((request) async {
      print(request.body);
      return Response('<root><SendMessageResult><foo></foo></SendMessageResult></root>', 200);
    });

    final sqs = SQS(
      region: 'foo',
      credentials: AwsClientCredentials(accessKey: 'null', secretKey: 'null'),
      client: client,
    );
    await sqs.sendMessage(
        messageBody: '{"foo":6}',
        queueUrl: 'http://foo.com',
        messageGroupId: 'bar',
        messageDeduplicationId: '666',
        messageAttributes: {
          'companyId': MessageAttributeValue(
              dataType: 'string', stringValue: 'companyId')
        });
  });
}
```
The request body of this is:
`Action=SendMessage&MessageAttributes.entry.1.key=companyId&MessageAttributes.entry.1.value.entry.1.key=DataType&MessageAttributes.entry.1.value.entry.1.value=string&MessageAttributes.entry.1.value.entry.2.key=StringValue&MessageAttributes.entry.1.value.entry.2.value=companyId&MessageBody=%7B%22test%22%3A+%22cool%22%7D&MessageDeduplicationId=4a7f57c07cf611eaf1cde9f670f7ada1&MessageGroupId=bar&QueueUrl={sqsUrl}&Version=2012-11-05`

Which differs a bit from what @lucasdina says is a valid:
`Action=SendMessage&MessageAttribute.1.Name=testattrib&MessageAttribute.1.Value.DataType=String&MessageAttribute.1.Value.StringValue=test&MessageBody=%7B%22test%22%3A+%22cool%22%7D&MessageDeduplicationId=4a7f57c07cf611eaf1cde9f670f7ada1&MessageGroupId=AgriPayFarms&QueueUrl={sqsUrl}&Version=2012-11-05`

shared_aws_api/lib/src/protocol/query.dart
```diff
@@ -119,6 +119,10 @@ Iterable<MapEntry<String, String>> _flatten(
     return;
   }
 
+  if (data is! Map) {
+    data = data.toJson();
```
isoos, line 123: I'm working on a patch that does not need this, maybe remove it for the time being?

generator/lib/model/api.dart
```diff
@@ -79,9 +87,9 @@ class Api {
 
   String get packageBaseName {
     final candidates = <String>[
-      metadata.endpointPrefix,
       metadata.uid?.split('-20')?.first,
       metadata.className.toLowerCase(),
+      metadata.endpointPrefix
```
isoos, line 92: Why this change? What's the impact?

shared_aws_api/lib/src/protocol/query.dart
isoos: On a second thought this can be here, just add a // TODO: to remove it once we call `.toJson` at request building time.

generator/lib/model/api.dart
Schwusch: RDS, DocDB and Neptune share the same endpoint prefix, hence they're all int the [RDS library](https://pub.dev/documentation/aws_rds_api/latest/index.html).
This change means they will be landing in separate packages.

PR #147, opened 2020-04-13
Updated flattening in shared query protocol.

isoos:
It seems that #144 requires a breaking change in the protocol library. I've a related change in the code generator that would update how the attribute's toJson() is called, and together with this PR they seem to generate the required format for `SQS`.

### DIFF
--- a/shared_aws_api/lib/src/protocol/query.dart
+++ b/shared_aws_api/lib/src/protocol/query.dart
@@ -72,6 +72,7 @@ class QueryProtocol {
       Map<String, dynamic> data, String method, String requestUri) {
     final rq = Request(method, Uri.parse('$_endpointUrl$requestUri'));
     rq.body = _canonical(flatQueryParams(data));
+    print(rq.body);
     rq.headers['Content-Type'] = 'application/x-www-form-urlencoded';
     // TODO: handle if the API is using different signing
     signAws4HmacSha256(
@@ -89,8 +90,8 @@ Map<String, String> flatQueryParams(dynamic data) {
   return Map.fromEntries(_flatten([], data));
 }
 
-Iterable<MapEntry<String, String>> _flatten(
-    List<String> prefixes, dynamic data) sync* {
+Iterable<MapEntry<String, String>> _flatten(List<String> prefixes, dynamic data,
+    {bool flat}) sync* {
   if (data == null) {
     return;
   }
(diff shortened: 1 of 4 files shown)


Schwusch:
I don't know if this helps, but there is an attribute `Shape.flattened` and `Member.flattened` that are true in [`MessageBodyAttributeMap`](https://github.com/aws/aws-sdk-js/blob/59512e1eba/apis/sqs-2012-11-05.normal.json#L971) for example. Perhaps that is useful information in this case?

isoos:
> I don't know if this helps, but there is an attribute Shape.flattened and Member.flattened that are true in MessageBodyAttributeMap for example. Perhaps that is useful information in this case?

It is very likely helpful, but how should we expose it?

Schwusch:
I am not following [the JS implementation very well](https://github.com/aws/aws-sdk-js/blob/59512e1eba/lib/query/query_param_serializer.js#L32), but the `flattened` parameter is passed around indirectly through the metadata

shared_aws_api/lib/src/protocol/query.dart
```diff
@@ -119,23 +119,24 @@ Iterable<MapEntry<String, String>> _flatten(
     return;
   }
 
-  // TODO: to remove it once we call .toJson at request building time.
-  if (data is! Map) {
-    data = data.toJson();
```
Schwusch, line 124: Where is classes serialized after this change?

isoos: I have a separate PR in the works for it. Can't make it part of this one, as it will change the generated code, and with that I'd like to upgrade the dependency, and that requires prior publishing of the shared library.

isoos:
> I am not following the JS implementation very well, but the flattened parameter is passed around indirectly through the metadata somehow.

They are passing the API + shape information as part of the argument flattening. For us it would be slightly different, eg. we could add a '__flattened': true, key as part of the `Map` and work with it accordingly? Not sure what else we can do with our two-phase approach.

Schwusch:
We could do the flattening in the generated classes, that is where we can put that kind of metadata.
That would mean sending the already flattened map to the `_protocol.send()` method.
The flattening algorithm is ~50 lines, which doesn't add much. WDYT?

isoos:
> We could do the flattening in the generated classes, that is where we can put that kind of metadata.
> That would mean sending the already flattened map to the _protocol.send() method.
> The flattening algorithm is ~50 lines, which doesn't add much. WDYT?

That is a good point, I'll give it a try (in the following days). I'll suspend this PR and close once we have the codegen available.

isoos:
Turns out this is wrong for many different reasons.

PR #148, opened 2020-04-13
Optional part for JSON serialization.

isoos:
`importexport` does not have any requirement for `toJson` or `fromJson`. We could probably do better and skip the annotations too, but it doesn't seem to have urgency other than not generating the `part 'importexport.g.dart';` block.


isoos:
> We could probably traverse all api operations and their input/output shapes before generating code, to build up maps that can tell us if a shape is used as e.g. input.

+1, that is a nice-to-have fix later...

PR #149, opened 2020-04-16
The generator code failed on the OutputS3KeyPrefix pattern

tyrchen:
In AWS Polly API definition, there's a pattern can be generated corrected:

```json
    "OutputS3KeyPrefix": {
      "type": "string",
      "pattern": "^[0-9a-zA-Z\\/\\!\\-_\\.\\*\\'\\(\\)]{0,800}$"
    }
```
We shall take care of this corner case.

### DIFF
--- a/generator/lib/builders/library_builder.dart
+++ b/generator/lib/builders/library_builder.dart
@@ -46,7 +46,7 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 import 'package:shared_aws_api/shared.dart' as _s;
-import 'package:shared_aws_api/shared.dart' 
+import 'package:shared_aws_api/shared.dart'
   show Uint8ListConverter, Uint8ListListConverter ${api.generateJson ? ', rfc822fromJson, rfc822toJson, iso8601fromJson, iso8601toJson, unixFromJson, unixToJson' : ''};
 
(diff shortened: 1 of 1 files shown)


isoos:
@tyrchen: I think this is a good fix, but let's @Schwusch also accept it. Also: please make sure `dartfmt -n .` does not report unformatted files and `dartanalyzer .` does not report any issues (otherwise CI is not happy, and we can't merge).

generator/lib/builders/library_builder.dart
```diff
@@ -140,7 +140,12 @@ ${builder.constructor()}
       final pattern = member.shapeClass?.pattern;
 
       if (pattern != null) {
-        writeln("_s.validateStringPattern('$name', $name, r'$pattern',);");
```
Schwusch, line 143: I say, why not future proof this and do `r'''$pattern'''`?

tyrchen: Thanks. This is much better. I changed the code. This fix will infect a lot of generated code, so I will add them as a separate commit.

tyrchen:
@isoos thanks for your suggestion! Now the code passed the CI.

Schwusch:
LGTM :)

PR #156, opened 2020-04-20
Updating generated RegExp pattern + latest API definitions. [patch]

isoos:
Unfortunately I've deleted my old API versions, and the two changes got mixed together. I'll try to be a better next time and will update the generated code right after a related PR lands...


Schwusch:
I believe there is a mistake in the code generation.
 If an operation argument is not required, there should be a null check before matching against regex(or any other check).
This is something I found in `SecretsManager.getSecret` but forgot until now.

isoos:
> I believe there is a mistake in the code generation.

It is not a regression, is it?

Schwusch:
No it is not, it just made some packages unusable. Some parameters, when left out, has a default behaviour which is not possible when the code crashes because they are left out.

PR #157, opened 2020-04-20
Preparing new release with a version bump for generated packages

isoos:


isoos:
@Schwusch FYI (will release them after merge)

PR #158, opened 2020-04-21
JSON protocol: enabled for publishing, generated packages.

isoos:
- added `for` as reserved name
- added packages to `config.yaml`
- added generated code to repository


Schwusch:
I'd really like if you considered https://github.com/agilord/aws_client/pull/159 before this change. It will mitigate a lot of pain for our users.

On another note, how long did these packages take for you to generate? :grinning:

isoos:
> On another note, how long did these packages take for you to generate?

I think it was about two hours.

isoos:
Let's land this 🍾

PR #161, opened 2020-04-21
Conditional validation when parameter is not required. [patch]

isoos:
Applying #159 on the generated packages. We could probably gain an extra line per check if e.g. `validateStringLength` could take an optional parameter `isRequired` and do the null check inside, maybe if we need to add another check for some reason...

### DIFF
--- a/generated/aws_elasticache_api/lib/elasticache-2015-02-02.dart
+++ b/generated/aws_elasticache_api/lib/elasticache-2015-02-02.dart
@@ -160,6 +160,8 @@ class ElastiCache {
     List<String> replicationGroupIds,
   }) async {
     ArgumentError.checkNotNull(serviceUpdateName, 'serviceUpdateName');
+    if (cacheClusterIds != null) {}
+    if (replicationGroupIds != null) {}
     final $request = <String, dynamic>{
       'Action': 'BatchApplyUpdateAction',
       'Version': '2015-02-02',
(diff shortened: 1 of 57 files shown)


generated/aws_elasticache_api/lib/elasticache-2015-02-02.dart
```diff
@@ -3316,6 +3321,10 @@ class ElastiCache {
     bool showNodeLevelUpdateStatus,
     List<String> updateActionStatus,
   }) async {
+    if (cacheClusterIds != null) {}
+    if (replicationGroupIds != null) {}
+    if (serviceUpdateStatus != null) {}
+    if (updateActionStatus != null) {}
     final $request = <String, dynamic>{
```
Schwusch, line 3328: This didn't go as expected..?

isoos: Good catch, maybe we should fix this before commiting the change... Are you up for it?

Schwusch: Yeah I'll take a shot at it

Schwusch: https://github.com/agilord/aws_client/pull/162

isoos: Thank you for the quick turnaround! I'm running the codegen again for both PRs.

PR #162, opened 2020-04-21
Handle conditional validation when parameter is not required

Schwusch:
Handle conditional validation when parameter is not required in shared library


isoos:
> bloody formatting

:D

Issue #160, opened 2020-04-21
Make AwsClientCredentials constructor argument optional


Schwusch:
In the JS SDK the usage is more ergonomic:
```javascript
const secretsManager = new AWS.SecretsManager({ 'region':  'eu-west-1' });
...
```
This step:
```dart
final environment = Platform.environment;
final credentials = AwsClientCredentials(
  accessKey: environment['AWS_ACCESS_KEY_ID'],
  secretKey: environment['AWS_SECRET_ACCESS_KEY'],
  sessionToken: environment['AWS_SESSION_TOKEN'],
);
// final secretsManager = SecretsManager(region: region, credentials: credentials);
```
We should be able to do that step in the protocol factory constructor if credentials are null.
That change would make onboarding new users smoother.

isoos:
sgtm

Schwusch:
The pattern for doing so in other AWS SDKs is described here, just disregard the Java specific stuff:
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html

isoos:
The only issue I have with the auto-detection is that right now the library is generic, it can be accessed on all the platforms. If we do auto-detection, it will to an extent, be hard-wired to use `dart:io`, making the libraries incompatible with JS apps.

Maybe a few indirection here and there would solve that... not sure yet what's the best decision here.

Schwusch:
We do conditional imports with stubs on the web, as demonstrated in this article: https://medium.com/@rody.davis.jr/how-to-build-a-native-cross-platform-project-with-flutter-372b9e4b504f

isoos:
Yeah, as long as `package:http` is using it, we can use it too...

PR #166, opened 2020-04-23
Resolve AwsClientCredentials if not provided

Schwusch:
Partially resolves https://github.com/agilord/aws_client/issues/160.
Tries to resolve credentials from these places if no credentials are provided:
- Environment Variables
- Credential profiles file at the default location (~/.aws/credentials) or `AWS_SHARED_CREDENTIALS_FILE` if set

Does not resolve from these sources:
- Credentials delivered through the Amazon EC2 container service if AWS_CONTAINER_CREDENTIALS_RELATIVE_URI" environment variable is set and security manager has permission to access the variable,
- Instance profile credentials delivered through the Amazon EC2 metadata service
- Web Identity Token credentials from the environment or container.


isoos:
Looks good! Please fix formatting.

Not urgent, but maybe in the future `resolve()` could also use a pre-populated value too.

Schwusch:
> use a pre-populated value too

Do you mean partial credentials, filling in the rest? Would it override or be overridden?

isoos:
> Do you mean partial credentials, filling in the rest? Would it override or be overridden?

`resolve()` is almost like a singleton. If it gets a "setter" (we can call it `setGlobalCredential` or `setResolveInstance`) then an app can set it up as a "global" variable and downstream code can just use it.

Schwusch:
> If it gets a "setter" then an app can set it up as a "global" variable 

I'm not certain how that would look, so I didn't add it this time.

isoos:
> I'm not certain how that would look, so I didn't add it this time.

Yeah, this can be figured out later.

PR #168, opened 2020-04-24
AwsClientCredentials is no longer a required parameter. [patch]

isoos:


isoos:
@Schwusch: how urgent is it for you to have these published?

Schwusch:
It is not urgent, no need to stress something out.
Are you planning to bump the versions in another PR?

isoos:
> Are you planning to bump the versions in another PR?

Yeah, but if it is not urgent to publish these, I may wait for it - we may have a few other things to be done over the weekend...

PR #170, opened 2020-04-26
Export XmlFindExtension in shared lib.

### LINKED ISSUE #169: XML based packages are broken
After the XML package was bumped in https://github.com/agilord/aws_client/pull/166/ `query`/`xml` packages broke. We either need to revert to <4.0.0 or update the usage.

isoos:
I think this fixes #169 (at least the packages I've checked locally).


isoos:
published

PR #171, opened 2020-04-26
Document client

Schwusch:
This is a work in progress, but I wanted you too have a say in the structure as early as possible.
This PR is basically for discussion about further development.

### DIFF
--- a/document_client/lib/document_client.dart
+++ b/document_client/lib/document_client.dart
@@ -1,0 +1,111 @@
+import 'package:aws_dynamodb_api/dynamodb-2012-08-10.dart';
+import 'package:shared_aws_api/shared.dart';
+import 'package:document_client/src/translator.dart';
+
+export 'package:aws_dynamodb_api/dynamodb-2012-08-10.dart';
+
+/// The document client simplifies working with items in Amazon DynamoDB
+/// by abstracting away the notion of attribute values. This abstraction
+/// annotates native Dart types supplied as input parameters, as well
+/// as converts annotated response data to native Dart types.
+///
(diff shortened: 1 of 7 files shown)


document_client/lib/document_client.dart
```diff
@@ -0,0 +1,111 @@
...
+  ///
+  ///  await documentClient.get(
+  ///   tableName: 'Table',
+  ///   key: {
+  ///     'HashKey': 'hashkey'
+  ///   }
+  ///  );
+  ///
+  Future<GetResponse> get({
+    @required String tableName,
```
isoos, line 83: An idea:
Do you anticipate other table-related methods to use the `@required String tableName`? Maybe it could be moved to a `TableClient` which takes it in the constructor and now you'll have one less argument for all these tables. `DocumentClient` may have a `TableClient table(String name)` method on it.

Schwusch: I'd say there is equal or more [cross-table operations than on a singular table](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB/DocumentClient.html), but it is an interesting abstraction for sure. 
I haven't used DynamoDB that extensively to say if it would be useful. 

Is it something you would benefit from?

isoos: I don't use DynamoDB, it just occurred to me as a possibility, but I have nothing against the current PR. Should I just merge it?

Schwusch: > Should I just merge it?

I tried it, it works, so we might as well. I can do new PRs when developing the rest.

PR #173, opened 2020-04-27
Nicer output when generating packages

Schwusch:
I grew tired of the output when generating packages, so I made it more informative.
I put some stuff in its place:
 - loading bar
 - percentage
 - elapsed time
 - estimated time left


Schwusch:
What it looks like:

What it looks like

PR #174, opened 2020-04-27
Generate tests to verify that generated code compiles

Schwusch:
Simple tests to ensure the generated code compiles.

### DIFF
--- a/generator/lib/builders/test_builder.dart
+++ b/generator/lib/builders/test_builder.dart
@@ -1,4 +1,4 @@
-String buildTest(String packageRelativeDirectory) =>
+String buildVerifyTest(String packageRelativeDirectory) =>
     '''@Tags(['presubmit-only'])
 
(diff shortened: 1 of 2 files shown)


isoos:
- Would this actually run in the absence of AWS keys?
- Wouldn't running `dartanalyzer` be better? (the recent failures would have been caught if we did run `pub upgrade` instead of `pub get` and checked the output of `dartanalyzer`.

Schwusch:
> Would this actually run in the absence of AWS keys?

The constructors only check for null, so it does run as long as no operations are called.

> Wouldn't running dartanalyzer be better?

You're probably right. I am trying to think how automated tests can be set up, like the lint we have now. Where should Dart analyzer be run?

isoos:
> Where should Dart analyzer be run?

I think this should be probably a separate command, where we run checks like a CI would run them.

Issue #175, opened 2020-05-19
AWS dynamodb Bug for ReturnConsumedCapacity


pY4x3g:
Hi,

thank you for your library. The aws_dynamodb_api is already working really fine.

I think I found a small bug, the ReturnConsumedCapacity is not working. When I use the option ```returnConsumedCapacity: ReturnConsumedCapacity.total``` for a query or an update operation (DynamoDB.updateItem/DynamoDB.query) I get the following errors.

for a query:
```
Unhandled exception:
Converting object to an encodable object failed: Instance of 'ReturnConsumedCapacity'
#0      _JsonStringifier.writeObject (dart:convert/json.dart:659:7)
#1      _JsonStringifier.writeMap (dart:convert/json.dart:740:7)
#2      _JsonStringifier.writeJsonValue (dart:convert/json.dart:695:21)
#3      _JsonStringifier.writeObject (dart:convert/json.dart:650:9)
#4      _JsonStringStringifier.printOn (dart:convert/json.dart:846:17)
#5      _JsonStringStringifier.stringify (dart:convert/json.dart:831:5)
#6      JsonEncoder.convert (dart:convert/json.dart:260:30)
#7      JsonCodec.encode (dart:convert/json.dart:168:45)
#8      JsonProtocol.send (package:shared_aws_api/src/protocol/json.dart:53:45)
#9      DynamoDB.query (package:aws_dynamodb_api/dynamodb-2012-08-10.dart:3006:42)
```
for an update:
```
Unhandled exception:
Converting object to an encodable object failed: Instance of 'ReturnConsumedCapacity'
#0      _JsonStringifier.writeObject (dart:convert/json.dart:659:7)
#1      _JsonStringifier.writeMap (dart:convert/json.dart:740:7)
#2      _JsonStringifier.writeJsonValue (dart:convert/json.dart:695:21)
#3      _JsonStringifier.writeObject (dart:convert/json.dart:650:9)
#4      _JsonStringStringifier.printOn (dart:convert/json.dart:846:17)
#5      _JsonStringStringifier.stringify (dart:convert/json.dart:831:5)
#6      JsonEncoder.convert (dart:convert/json.dart:260:30)
#7      JsonCodec.encode (dart:convert/json.dart:168:45)
#8      JsonProtocol.send (package:shared_aws_api/src/protocol/json.dart:53:45)
#9      DynamoDB.updateItem (package:aws_dynamodb_api/dynamodb-2012-08-10.dart:4679:42)
```

[...]

PR #176, opened 2020-05-20
[patch] fix enum serialization in methods bug

Schwusch:
This corrects a bug reported in #175, where enum arguments in methods does not serialize correctly.


isoos:
Nice!

PR #177, opened 2020-05-21
Fixed top-level enum serialization. [patch]

isoos:
Generated code for #176.


isoos:
@Schwusch: I'm wondering: is `.toValue()` the correct call here? Shouldn't we call `?.toValue()` instead?

Schwusch:
> Shouldn't we call ?.toValue() instead?

You're right, there's probably no harm in doing it for required parameters too.
I can update the PR.

isoos:
> I can update the PR.

Yeah, please push a commit, I'll update the generated code.

isoos:
You can push a commit into this PR too - I think... But I can edit the code, no problem...

isoos:
To be clear, I'll do it.

PR #183, opened 2020-05-23
Add batchWrite() to DocumentClient

Schwusch:
There is a type bug in the `Uint8ListListConverter` which will break `AttributeValue.bs` decoding until a new version of `shared_aws_api` with the d3c0bce fix int this PR is in and released. 

DynamoDB will then have to be rebuilt with `build_runner` and released, depending on the new `shared_aws_api` version.

I've also added tests (that's how I found the bug ^^).

### DIFF
--- a/document_client/lib/document_client.dart
+++ b/document_client/lib/document_client.dart
@@ -145,13 +145,13 @@ class DocumentClient {
 
     return BatchGetResponse(
       response.consumedCapacity,
-      response.responses.map(
+      response.responses?.map(
         (k, v) => MapEntry(
           k,
           v.map((e) => e.toJson()).toList(),
         ),
       ),
-      response.unprocessedKeys.map(
+      response.unprocessedKeys?.map(
         (k, v) => MapEntry(
           k,
           KeysAndProjection(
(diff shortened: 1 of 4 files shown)


isoos:
> DynamoDB will then have to be rebuilt with build_runner and released, depending on the new shared_aws_api version.

Let me know when you want me to do this (or if you have other changes lined up).

isoos:
> DynamoDB will then have to be rebuilt with build_runner

@Schwusch: maybe I'm missing something, but there was no change in my build in `generated/aws_dynamodb_api`, and I don't see why there should be any. I think we can just release the shared library and advise people to run `pub upgrade`?

Schwusch:
@isoos When overriding in `aws_dynamodb_api/pubspec.yaml` with:
```yaml
dependency_overrides:
  shared_aws_api:
    path: ../../shared_aws_api
```
After running `pub run build_runner build` in `aws_dynamodb_api` I get:
image

That is the changes that fixes the crash. The shared library can and perhaps should be released before :)

isoos:
published

PR #195, opened 2020-05-28
[DocumentClient] added support for AttributeValues for the translator

### LINKED ISSUE #194: [DocumentClient] How to add a List of Numbers (not a NumberSet)?
Hi,
how do I add a List of Numbers to the database? The translator is translating any number list to a NumberSet which is not ordered and completely random in the dynamodb (`translator.dart:19`). Maybe we should check if the value is already an AttributeValue and return it in this case in the `toAttributeValue` method? With this I could just create the AttributeValue with a value `l` of a list of AttributeValues with type number like
```dart
  List<AttributeValue> dataListToAdd = [];

  for (var myNumber in [1,2,3]) {
    dataListToAdd.add(AttributeValue(n: myNumber.toString()));
  }
  var myValue = AttributeValue(l: dataListToAdd);
```

This was how it works with the DynamoDB class before. But currently this will create a json object and this will store it as a Map.
But that's only my first idea, maybe there is something better. Because it seems a bit random if the library uses a list or a set but this makes a huge difference for the user of the library.

Edit: I tried the simple fix by adding
```dart
  } else if (value is AttributeValue) {
    return value;
  }
```
to the `toAttributeValue` method and it is working for me.

pY4x3g:
This should fix #194

### DIFF
--- a/document_client/lib/src/translator.dart
+++ b/document_client/lib/src/translator.dart
@@ -5,6 +5,8 @@ import 'package:aws_dynamodb_api/dynamodb-2012-08-10.dart';
 AttributeValue toAttributeValue(dynamic value) {
   if (value == null) {
     return AttributeValue(nullValue: true);
+  } else if (value is AttributeValue) {
+    return value;
   } else if (value is bool) {
     return AttributeValue(boolValue: value);
   } else if (value is String) {


Schwusch:
LGTM

PR #201, opened 2020-05-29
Use pub upgrade instead of pub get

Schwusch:
When developing locally, the lockfile can create some discrepancies compared to Travis builds.
Warnings e.g. deprecations pops up in Travis and fails the build, so this aims to lessen those discrepancies.

### DIFF
--- a/generator/bin/generate.dart
+++ b/generator/bin/generate.dart
@@ -1,11 +1,13 @@
 import 'package:args/command_runner.dart';
-
 import 'package:aws_client.generator/download_command.dart';
 import 'package:aws_client.generator/generate_command.dart';
 import 'package:aws_client.generator/package_command.dart';
 
 void main(List<String> arguments) async {
-  final runner = CommandRunner('generator', 'AWS API generator tool')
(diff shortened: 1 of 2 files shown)


isoos:
> it seems the commands are similar in all ways except upgrade seems to not care about the lockfile, am I correct?

`upgrade` actively tries to use the latest versions, while `get` keeps the currently resolved ones, unless there is a stricter (new) constraint in the `pubspec.yaml` file.

isoos:
I see uses for both `get` and `upgrade` (based on the knowledge that during development we do know which we want). Maybe this should be a flag?

Schwusch:
I can add it as a flag. I'm thinking that `upgrade` result in less Travis errors, and thus should be defaults?

isoos:
> I can add it as a flag. I'm thinking that upgrade result in less Travis errors, and thus should be default?

sgtm

Schwusch:
I added a flag to skip the git check optimization as well, since the build step can depend on changes in the shared library as well.

Issue #172, opened 2020-04-27
SES NoSuchMethodError: Class 'Destination' has no instance method 'toJson'.


latesr:
Whilst trying out your SES support, I encountered the following exception.

```Unhandled exception:
NoSuchMethodError: Class 'Destination' has no instance method 'toJson'.
Receiver: Instance of 'Destination'
Tried calling: toJson()
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:53:5)
#1      _flatten (package:shared_aws_api/src/protocol/query.dart:124:17)
#2      _SyncIterator.moveNext (dart:core-patch/core_patch.dart:146:12)
#3      _SyncIterator.moveNext (dart:core-patch/core_patch.dart:139:32)
#4      MapMixin.addEntries (dart:collection/maps.dart:172:23)
#5      new Map.fromEntries (dart:core/map.dart:187:17)
#6      flatQueryParams (package:shared_aws_api/src/protocol/query.dart:89:14)
#7      QueryProtocol._buildRequest (package:shared_aws_api/src/protocol/query.dart:74:26)
#8      QueryProtocol.send (package:shared_aws_api/src/protocol/query.dart:49:16)
#9      SES.sendEmail (package:aws_ses_api/email-2010-12-01.dart:2208:37)
```

The following code demonstrates this minimally:

```dart 
import 'package:aws_ses_api/email-2010-12-01.dart';

void main(List<String> arguments) async {
  final cred = AwsClientCredentials(
      accessKey: '',
      secretKey: '');
  final ses = SES(region: 'eu-west-1', credentials: cred);
  await ses.sendEmail(
      destination: Destination(toAddresses: ['dest@example.com']),
      message: Message(
        body: Body(
          text: Content(data: 'My test text'),
        ),
        subject: Content(data: 'Test subject'),
      ),
      source: 'source@example.com');
}
```

Schwusch:
This can probably be solved at the same time as https://github.com/agilord/aws_client/issues/144. 

@isoos Are you doing any work in this?

isoos:
> @isoos Are you doing any work in this?

Not at the moment :(

Schwusch:
This should be resolved in [version 0.1.1](https://pub.dev/packages/aws_ses_api/versions/0.1.1). @latesr can you verify?

Schwusch:
I'll close this for now. The activity indicates that this may not a problem anymore.

PR #204, opened 2020-06-01
Generate toJson() for parameter classes in query protocol

Schwusch:
I tried the example in #172 with a mock client, which didn't crash. So I got that going atleast.
It might be a step towards solving #144 as well, maybe.

### DIFF
--- a/generator/lib/builders/library_builder.dart
+++ b/generator/lib/builders/library_builder.dart
@@ -221,11 +221,11 @@ ${builder.constructor()}
       if (shape.deprecated) {
         writeln(r'@deprecated');
       }
-      if (shape.api.generateJson) {
+      if (shape.requiresJson) {
         writeln(
             '@_s.JsonSerializable(includeIfNull: false, explicitToJson: true, '
-            'createFactory: ${shape.isUsedInOutput && shape.api.generateFromJson},'
-            'createToJson: ${shape.isUsedInInput && shape.api.generateToJson})');
+            'createFactory: ${shape.generateFromJson},'
+            'createToJson: ${shape.generateToJson})');
       }
 
       final extendsBlock = shape.exception ? 'implements _s.AwsException ' : '';
@@ -239,7 +239,7 @@ ${builder.constructor()}
         final shapeClass = member.shapeClass;
         final valueEnum = shapeClass.enumeration;
 
-        if (shape.api.generateJson) {
+        if (shape.requiresJson) {
           var dateTimeConversion = '';
 
           if (member.dartType == 'Uint8List') {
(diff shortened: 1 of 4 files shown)


generator/lib/builders/library_builder.dart
```diff
@@ -276,12 +276,12 @@ ${builder.constructor()}
         write('\n  $name({${constructorMembers.join()}});');
       }
 
-      if (shape.api.generateFromJson && shape.isUsedInOutput) {
```
isoos, line 279: I'd rather keep using `isUsedInOutput` and `isUsedInInput` (it may help to reduce the code size + build time), even if we set it to true for now in this protocol.

generator/lib/builders/library_builder.dart
```diff
@@ -221,11 +221,11 @@ ${builder.constructor()}
       if (shape.deprecated) {
         writeln(r'@deprecated');
       }
-      if (shape.api.generateJson) {
+      if (shape.requiresJson) {
```
isoos, line 224: Not sure: what's changing here?

Schwusch: The difference is that each shape decides whether it needs JSON serialisation, rather than the API as a whole.
The `query` protocol shouldn't have JSON serialisation at all on output classes, but all output classes are marked as used in [putOperation](https://github.com/agilord/aws_client/blob/55da9343639aa6dd165707e8b3f1ec94ed531215/generator/lib/builders/library_builder.dart#L109)
The result of `shape.api.generateJson == true` when the shape is used in output and ´shape.api.generateFromJson == false` is:
```dart
@_s.JsonSerializable(
  includeIfNull: false, 
  explicitToJson: true,
  createFactory: false,
  createToJson: false,
)
```
Which is useless and won't generate anything of value. When asking `shape.requiresJson`, the shape will only go into the `if`-statement when it is used in either input or output and really should generate JSON-serialisation.

generator/lib/builders/library_builder.dart
Schwusch: >  (it may help to reduce the code size + build time)

I'm not sure what you mean.
The logic is the same after the change, but is [contained in the `shape`](https://github.com/agilord/aws_client/blob/55da9343639aa6dd165707e8b3f1ec94ed531215/generator/lib/model/shape.dart#L152). I partly explained it in the [other answer](https://github.com/agilord/aws_client/pull/204/files/55da9343639aa6dd165707e8b3f1ec94ed531215#r433688994), but the `shape` has more potential to answer if JSON logic should be generated, especially in protocols like `query`. `query` is a special case, and there will be more special cases once `ec2` and `rest-json` is worked on.

isoos: Ah, sorry, good point.

Issue #138, opened 2020-03-03
Generate tests


Schwusch:
We should be able to generate protocol test cases, much like [aws-sdk-go](https://github.com/aws/aws-sdk-go-v2/tree/master/models/protocol_tests). 
It would help development a lot to know if the implementations are correct.

isoos:
Nice idea!

Schwusch:
The Go SDK generate [smoke tests for the APIs as well](https://github.com/aws/aws-sdk-go-v2/blob/master/private/model/api/smoke.go), I am investigating if it is feasible in our case as well, using their test definitions.

Schwusch:
I see now that [aws-sdk-has the same protocol tests,](https://github.com/aws/aws-sdk-js/tree/master/test/fixtures/protocol) might be good to know

Schwusch:
@xvrh solved this for us :D

PR #208, opened 2020-06-14
Adds metadata to generated query services and test cases for query and json protocol

Schwusch:
This is a WIP, but I wanted to show how I thought this could work.
I implemented the Go tests mentioned in #138 to make testing more accurate.
This is intended to solve #144.


Schwusch:
@isoos as you can see, now Travis fails because of a unit test (our faulty implementation) :smile: 
Now we can implement it the right way, but with test driven development :partying_face:

PR #217, opened 2020-07-27
Headers are not sent with a RestXmlRequest

thomas94100:
Hello, 

The headers map passed to the send() function of the RestXmlProtocol is not passed to the _buildRequest() method and are not sent to the server. As a result, S3 object metadatas cannot be uploaded.

### DIFF
--- a/shared_aws_api/lib/src/protocol/rest-xml.dart
+++ b/shared_aws_api/lib/src/protocol/rest-xml.dart
@@ -63,9 +63,7 @@ class RestXmlProtocol {
       final code = error.findElements('Code').first.text;
       final message = error.findElements('Message').first.text;
       final fn = exceptionFnMap[code];
-      final exception = fn != null
-          ? fn(type, message)
-          : GenericAwsException(type: type, code: code, message: message);
+      final exception = fn != null ? fn(type, message) : GenericAwsException(type: type, code: code, message: message);
       throw exception;
     }
     if (resultWrapper != null) {
@@ -74,13 +72,7 @@ class RestXmlProtocol {
     return RestXmlResponse(rs.headers, elem);
   }
 
-  Request _buildRequest(
-    String method,
-    String requestUri,
-    Map<String, String> queryParams,
-    dynamic payload,
-    Map<String,String> headers
-  ) {
+  Request _buildRequest(String method, String requestUri, Map<String, String> queryParams, dynamic payload, Map<String, String> headers) {
     queryParams ??= <String, String>{};
     final rq = Request(
       method,
@@ -88,7 +80,7 @@ class RestXmlProtocol {
         '$_endpointUrl$requestUri',
       ).replace(queryParameters: queryParams.isEmpty ? null : queryParams),
     );
-    if(headers !=null){
+    if (headers != null) {
       rq.headers.addAll(headers);
     }
     if (payload is XmlElement) {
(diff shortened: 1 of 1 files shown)


Schwusch:
LGTM, Travis seems to complain on the formatting. 
Please run `dartmft -w .` in the shared project.
@isoos wdyt?

isoos:
lgtm, after formatting is fixed.

thomas94100:
Sorry, it is my first pull request.

Cannot figure out what is wrong with my formatting, I run the VSCode dart plugin "format" command which should use dartfmt...

isoos:
what Dart SDK are you using?

thomas94100:
I am using the one that comes with Flutter : 
Here is the output of the ` flutter --version` command : 
`Flutter 1.17.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision b041144f83 (8 weeks ago) • 2020-06-04 09:26:11 -0700
Engine • revision ee76268252
Tools • Dart 2.8.4`

isoos:
> Dart 2.8.4

This looks good. So.. what happens if you run `dartfmt -w .` in the project library?

thomas94100:
`aws_client\shared_aws_api> dartfmt -w lib/
Formatting directory lib/:
Unchanged shared.dart
Unchanged src\credentials\credentials_html.dart
Unchanged src\credentials\credentials_io.dart
Unchanged src\credentials\ini_config.dart
Unchanged src\credentials.dart
Unchanged src\model\descriptor.dart
Unchanged src\model\descriptor.g.dart
Unchanged src\model\shape.dart
Unchanged src\model\shape.g.dart
Unchanged src\protocol\ec2.dart
Unchanged src\protocol\json.dart
Unchanged src\protocol\query.dart
Unchanged src\protocol\rest-json.dart
Formatted src\protocol\rest-xml.dart
Unchanged src\protocol\shared.dart
Unchanged src\protocol\_sign.dart
Unchanged src\scoping_extensions.dart
Unchanged src\validation\utils.dart`

isoos:
Thanks, I think the formatting is OK now. However, a few things in the JSON tests are failing:
https://travis-ci.com/github/agilord/aws_client/jobs/366044797

@Schwusch can you take a look if these are related to this PR?

Schwusch:
It is only my broken unit tests failing. Nothing related to this PR :+1:

isoos:
Do we have anything in the pipeline, or shall I just publish a new version with this?

Schwusch:
I don't have anything else, but I believe there is breaking changes in the shared library since my work on it.
Perhaps bump a major and release new versions on all libraries?

isoos:
Ah, what was the breaking change?

Schwusch:
I remember [adding 2 more parameters](https://github.com/agilord/aws_client/blob/master/shared_aws_api/lib/src/protocol/query.dart#L52-L53) to the Query protocol at least, which won't function without them. And then there is _a lot_ of bugfixes on all the protocols, so I think they all deserve to be promoted to 1.0.0.

isoos:
I wouldn't rush to `1.0.0` yet (but we shall do that once we are fully supporting every protocol and for a while there are no bug reports). Let's go with `0.2.0` for now...

Schwusch:
Ah, I just read up on how the [caret syntax](https://dart.dev/tools/pub/dependencies#caret-syntax) works, looks like it won't break in the `0.2.0` case 🎉

PR #221, opened 2020-08-05
Fix datetime converter errors, speed up build time

Schwusch:
I made some mistakes naming the date serializer functions, which results in [broken packages](https://pub.dev/packages/aws_dynamodb_api/versions/0.1.0/score)(all packages with date serialization basically).
This fixes that problem hopefully, and **most important of all** it reduced the code generation time from ~80 minutes to ~12 on my machine with 8 cores! :tada:

### DIFF
--- a/shared_aws_api/lib/src/protocol/shared.dart
+++ b/shared_aws_api/lib/src/protocol/shared.dart
@@ -8,15 +8,20 @@ import 'package:xml/xml.dart';
 final rfc822Formatter = DateFormat("EEE, dd MMM yyyy HH:mm:ss 'Z'", 'en_US');
 final iso8601Formatter = DateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", 'en_US');
 
-DateTime rfc822fromJson(String date) => rfc822Formatter.parse(date);
+DateTime rfc822FromJson(String date) =>
+    date == null ? null : rfc822Formatter.parse(date);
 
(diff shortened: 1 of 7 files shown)


shared_aws_api/lib/src/protocol/shared.dart
```diff
@@ -8,15 +8,20 @@ import 'package:xml/xml.dart';
 final rfc822Formatter = DateFormat("EEE, dd MMM yyyy HH:mm:ss 'Z'", 'en_US');
 final iso8601Formatter = DateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", 'en_US');
 
-DateTime rfc822fromJson(String date) => rfc822Formatter.parse(date);
+DateTime rfc822FromJson(String date) =>
```
isoos, line 11: This is technically a breaking change. Could you please also update the version and Changelog highlighting it?

Schwusch: Considering it was broken before, should this perhaps be a patch increment? I'll bump it to 0.2.1..

isoos: > Considering it was broken before, should this perhaps be a patch increment?

👀 :D

isoos:
And the packages... de they need to be re-generated and re-published?

Schwusch:
> re-generated and re-published?

They do, but it hopefully won't take all night with this change 😄  Depending on the number of cores of course.
Perhaps they should be depending on shared `^0.2.1`?

PR #228, opened 2020-08-18
rest-json tests and initial protocol implementation

Schwusch:


generator/lib/builders/protocols/service_builder.dart
```diff
@@ -21,8 +21,8 @@ abstract class ServiceBuilder {
     if (sc != null) {
       sc.uriMembers.forEach((m) {
         uri = uri
-            .replaceAll('{${m.locationName ?? m.name}}', '\$${m.fieldName}')
-            .replaceAll('{${m.locationName ?? m.name}+}', '\$${m.fieldName}');
+            .replaceAll('{${m.locationName ?? m.name}}', '\${Uri.encodeComponent(${m.fieldName}.toString())}')
```
isoos, line 24: nit: the `.toString()` is not required

isoos, line 24: Note: If the locationName component is in a full segment, maybe it could be safer to use `Uri.pathSegments` for that.

isoos (approved):
lgtm

generator/lib/builders/protocols/service_builder.dart
Schwusch: `Uri.encodeComponent` takes a `String`, and the argument can, and is, sometimes an `int` for example.
Although, You can make the argument that we only add `.toString()` when the type is not a string.

Schwusch: I blindly, and partially implemented what the comment above stated: 
`// TODO implement UriEscape and UriEscapePath according to https://github.com/aws/aws-sdk-js/blob/master/lib/util.js#L39..L57`.
I'm not sure what you mean by segments, should we do this after we constructed the path in its entirety?

isoos: Ah, true, leave the `.toString()` for now, no reason to optimize it...

isoos: Let's keep what you have. I thought that this is applied only on the `path` segment of the `uri`, but that assumption is not necessarily true.

Issue #206, opened 2020-06-03
putObject working but throws error


recepaliskan1:
Hello, s3 upload successful but throws error
return error message = ""<" expected"

My code:
`try {
          await service.putObject(
              bucket: bucketName,
              key: fileName,
              body: imageFile.readAsBytesSync(),
              acl: ObjectCannedACL.publicRead);
        } catch (ex) {
          var error= ex;
        }`

Schwusch:
Hi @recepaliskan1 !
Do you by chance have the stacktrace? 
I unfortunately don't have a possibility to reproduce this myself, 
so any more information would help a lot.

recepaliskan1:
putObject -> _protocol.send(...) -> parse(body) 
body is ""  
and parse method throw this error 

![git](https://user-images.githubusercontent.com/65152076/83721704-442ce000-a644-11ea-8e34-3fc98419a5fd.PNG)

Schwusch:
@isoos Does this ring a bell for you?

isoos:
@recepaliskan1: What happens if you call `getObject()` on a known, existing object in the bucket? Does it work?

Also: is there any chance that you can include the actual exception stacktrace?

recepaliskan1:
stacktrace:

```
I/flutter ( 2685): #0      parse 
package:xml/xml.dart:80
I/flutter ( 2685): #1      RestXmlProtocol.send 
package:shared_aws_api/…/protocol/rest-xml.dart:56
I/flutter ( 2685): <asynchronous suspension>
I/flutter ( 2685): #2      S3.putObject 
package:aws_s3_api/s3-2006-03-01.dart:7004
I/flutter ( 2685): #3      FileUploadViewState.uploadImage 
package:unideva_mobile/…/fileUpload/file_upload_view.dart:239
I/flutter ( 2685): <asynchronous suspension>
I/flutter ( 2685): #4      FileUploadViewState.showNameDialog.<anonymous closure> 
package:unideva_mobile/…/fileUpload/file_upload_view.dart:206
I/flutter ( 2685): #5      _InkResponseState._handleTap 
package:flutter/…/material/ink_well.dart:772
I/flutter ( 2685): #6      _InkResponseState.build.<anonymous closure> 
package:flutter/…/material/ink_well.dart:855
I/flutter ( 2685): #7      GestureRecognizer.invokeCallback 
package:flutter/…/gestures/recognizer.dart:182
[...]

PR #229, opened 2020-08-19
Add initial querystring implementation for rest-json

Schwusch:
I made a quick implementation for the querystring stringification.
I also added a CLI argument, so it's possible to generate all services with a specific protocol.

### DIFF
--- a/generator/lib/builders/protocols/rest_json_builder.dart
+++ b/generator/lib/builders/protocols/rest_json_builder.dart
@@ -20,13 +20,29 @@ class RestJsonServiceBuilder extends ServiceBuilder {
   @override
   String operationContent(Operation operation) {
     final buf = StringBuffer();
-    final shapeClass = operation.input?.shapeClass;
+    final inputShape = operation.input?.shapeClass;
+
     buildRequestHeaders(operation, buf);
-    if (!operation.http.bodyForbidden && shapeClass != null) {
-      final payload = shapeClass?.payloadMember;
+
+    buf.writeln("var query = '';");
+
+    if (inputShape?.hasQueryMembers == true) {
+      buf.writeln("query = '?\${[");
+      for (final member in inputShape.queryMembers) {
+        buf.writeln('if(${member.fieldName} != null)');
+        buf.writeln(
+            '_s.toQueryParam(${member.locationName == null ? 'null' : "'${member.locationName}'"}, ${member.fieldName}),');
+      }
+      buf.writeln("].where((e) => e != null).join('&')}';");
+    }
+
+    final outputClass = operation.output?.shapeClass?.className;
+
+    if (!operation.http.bodyForbidden && inputShape != null) {
+      final payload = inputShape?.payloadMember;
       if (payload == null) {
         buf.writeln('final \$payload = <String, dynamic>{');
-        shapeClass.members.where((m) => m.isBody).forEach((member) {
+        inputShape.members.where((m) => m.isBody).forEach((member) {
           var serializationSuffix = '';
           if (member.shapeClass.enumeration != null) {
             member.shapeClass.isTopLevelInputEnum = true;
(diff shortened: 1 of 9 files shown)


generator/lib/builders/protocols/rest_json_builder.dart
```diff
@@ -20,13 +20,29 @@ class RestJsonServiceBuilder extends ServiceBuilder {
...
     buildRequestHeaders(operation, buf);
-    if (!operation.http.bodyForbidden && shapeClass != null) {
-      final payload = shapeClass?.payloadMember;
+
+    buf.writeln("var query = '';");
+
+    if (inputShape?.hasQueryMembers == true) {
+      buf.writeln("query = '?\${[");
+      for (final member in inputShape.queryMembers) {
+        buf.writeln('if(${member.fieldName} != null)');
```
isoos, line 32: Let's generate full parenthesis for this  `if ([...]) {` ... `}`

generator/lib/generate_command.dart
```diff
@@ -141,7 +147,8 @@ in the config file, from the downloaded models.''';
         final protocolConfig = config.protocols[api.metadata.protocol];
         if (api.isRecognized &&
             (config.packages == null ||
-                config.packages.contains(api.packageName))) {
+                config.packages.contains(api.packageName)) && !(protocol != null && api.metadata.protocol != protocol)) {
```
isoos, line 150: This is getting nested... Maybe we should do `if (condition) continue;` in 2-3 lines?

shared_aws_api/lib/src/protocol/rest-json.dart
```diff
@@ -42,7 +42,7 @@ class RestJsonProtocol {
         client, service, region, endpointUrl, credentials);
   }
 
-  Future<JsonResponse> send({
+  Future<Map<String, dynamic>> send({
```
isoos, line 45: Why not use the typed `JsonResponse` anymore?

isoos (commented):
looks good, mostly nits

shared_aws_api/lib/src/protocol/rest-json.dart
Schwusch: I think I've seen somewhere that output shapes have the possibility to take information from the headers as well. 
An option would be to split them up and "manually" collect the information from the headers and inject it in the output shape.
That would require us to rethink the whole `ResponseClass.toJson()` approach though. As long as the header naming doesn't clash with naming in the body it should be fine I think.

generator/lib/generate_command.dart
Schwusch: I did that at first, the problem is the `else` clause, where the non generated services are registered. 
With the `if (condition) continue;` solution, the skipped services do not get registered.

generator/lib/builders/protocols/rest_json_builder.dart
Schwusch: I'm not sure I follow. This generates a "collection if" like described in the language tour:
```dart
var nav = [
  'Home',
  'Furniture',
  'Plants',
  if (promoActive) 'Outlet'
];
```
I don't think it works with curly brackets. Or did you mean something else?

shared_aws_api/lib/src/protocol/rest-json.dart
Schwusch: At lines 74-77, parsed body and header maps are merged.

Schwusch: [Here's an example of what I mean](https://github.com/agilord/aws_client/issues/206#issuecomment-677605530), we should fix this in the other protocols as well I think.

generator/lib/generate_command.dart
isoos: How about:
````dart
// this is the else clause as a function:

      void registerNotGenerated(Api api) {
          notGeneratedApis[api.metadata.protocol] ??= {};
          notGeneratedApis[api.metadata.protocol]
              [api.packageBaseName ?? 'NO_PACKAGE_BASENAME'] ??= [];
          notGeneratedApis[api.metadata.protocol]
                  [api.packageBaseName ?? 'NO_PACKAGE_BASENAME']
              .add(api.fileBasename);
      }
````
and then:
````dart
    if (protocol != null && api.metadata.protocol != protocol) {
      registerNotGenerated(api);
      continue;
    }
````
wdyt?

shared_aws_api/lib/src/protocol/rest-json.dart
isoos: ack

generator/lib/builders/protocols/rest_json_builder.dart
isoos: my bad, I was not looking at the context. I've got my daily coffee since :)

generator/lib/generate_command.dart
Schwusch: Yeah that is cleaner. I'll see what I can do.

Schwusch:
The diff is really going bananas when a large scope changes indentation...

isoos:
> The diff is really going bananas when a large scope changes indentation...

Yeah, I can see that :)

isoos (approved):
lgtm

PR #231, opened 2020-08-20
Json protocol

Schwusch:
This ensures that all data is returned as a `Map<String, dynamic>`, currently residing in `JsonResponse.body`.
This aims to be non breaking, but does not ensure that the services will be able to utilize the header data.
Header data is commonly referred to with its `locationName`, which has not been considered in the code generation thus far.


Schwusch:
@isoos This, together with #230 would be suitable for a patch release :)

isoos (approved):

PR #246, opened 2020-10-22
Use Dart CLI to run pub

xvrh:
- Use `dart pub` to run the pub commands. If you only have a flutter installation in your path, you don't have the old `pub`.
- Fix the `--optimize-build` flag: `--no-optimize-build` was doing nothing instead of re-building everything.
- Remove dead code
- Regenerate everything

### DIFF
--- a/generated/aws_redshift_api/lib/redshift-2012-12-01.g.dart
+++ b/generated/aws_redshift_api/lib/redshift-2012-12-01.g.dart
@@ -34,7 +34,7 @@ Map<String, dynamic> _$NodeConfigurationOptionsFilterToJson(
   writeNotNull(
       'Name', _$NodeConfigurationOptionsFilterNameEnumMap[instance.name]);
   writeNotNull('Operator', _$OperatorTypeEnumMap[instance.operator]);
-  writeNotNull('Values', instance.values);
+  writeNotNull('Value', instance.values);
   return val;
 }
 
--- a/generator/lib/generate_command.dart
+++ b/generator/lib/generate_command.dart
@@ -246,7 +246,7 @@ in the config file, from the downloaded models.''';
 
         touchedDirs.add(baseDir);
         generatedApis[api.packageName] = api.metadata.serviceFullName;
-        
+
         final pathParts = baseDir.split('/')..removeAt(0);
         final ensureBuildTestContent =
             formatter.format(buildTest(pathParts.join('/'), api));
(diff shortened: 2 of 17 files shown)


generator/lib/generate_command.dart
```diff
@@ -303,8 +303,9 @@ in the config file, from the downloaded models.''';
 
   Future<void> _getDependencies(String baseDir, {bool upgrade = true}) async {
     final pr = await Process.run(
-      'pub',
+      'dart',
```
Schwusch, line 306: thought: This is a feature available from Dart 2.10 and upwards. Perhaps we should bump the required SDK version in the generator in order to ensure that this is available?

xvrh: Good idea. I updated the pubspec.yaml

generated/aws_redshift_api/lib/redshift-2012-12-01.g.dart
```diff
@@ -34,7 +34,7 @@ Map<String, dynamic> _$NodeConfigurationOptionsFilterToJson(
   writeNotNull(
       'Name', _$NodeConfigurationOptionsFilterNameEnumMap[instance.name]);
   writeNotNull('Operator', _$OperatorTypeEnumMap[instance.operator]);
-  writeNotNull('Values', instance.values);
+  writeNotNull('Value', instance.values);
```
Schwusch, line 37: question: Is this the result of this PR code change, or could it be a different JSON api spec than originally generated with?
@isoos Have you generated this code with a clean slate, or might this be the result of old api definitions?

xvrh: I don't know what caused this change. 
I just run: `dart bin/generate.dart --download --no-optimize-build`

Should I specify the api version or the config file somewhere?

Or I can revert the changes in the generated folder.

Schwusch: The api version is defined in [config.yaml](https://github.com/agilord/aws_client/blob/master/generator/config.yaml#L1), but it is fully possible that the last generation was made with another version.
My opinion is that the build should be idempotent, so any differences are to be squashed :)
This justifies a new release for those services though. @isoos WDYT?

Schwusch: In the meanwhile, I want to merge this.
We can make a decision about releases later.

isoos: Yeah, it is something to look into, especially for changes like this. Maybe we should vendor the API definitions and store it in the repo too? It would make diffing explicit and straightforward...

PR #247, opened 2020-10-22
Use region_config_data.json to infer service url

### LINKED ISSUE #244: IAM api does not support regions
Hi, thank you for this awesome libraries :-)

The generated `IAM` api (`iam-2010-05-08.dart`) takes a region parameter and forwards it to the QueryProtocol to build the api endpoint.

From the AWS documentation (https://docs.aws.amazon.com/general/latest/gr/rande.html), the IAM API do not support region.

The IAM API fails with a `SocketException: Failed host lookup: 'iam.eu-west-1.amazonaws.com'` error

xvrh:
Fixes #244 

This is a first draft to use the correct service url for global services.

This PR:
- Saves the `region_config_data.json` file in the download phase
- Generates a file `shared_aws_api/lib/src/protocol/endpoint_config_data.dart` with the extracted data
- Create the `Endpoint` class that tries to infer the url and signing region based on the config data. (The code is taken from https://github.com/aws/aws-sdk-js/blob/master/lib/region_config.js)
- Modify the generator to not emit the `@required` annotation for supposed "global" services.


generator/lib/builders/protocols/rest_xml_builder.dart
```diff
@@ -10,9 +10,10 @@ class RestXmlServiceBuilder extends ServiceBuilder {
 
   @override
   String constructor() {
+    final regionRequired = api.isGlobalService ? '' : '@_s.required';
     return '''
     final _s.RestXmlProtocol _protocol;
-    ${api.metadata.className}({@_s.required String region, _s.AwsClientCredentials credentials, _s.Client client, String endpointUrl,})
+    ${api.metadata.className}({$regionRequired String region, _s.AwsClientCredentials credentials, _s.Client client, String endpointUrl,})
```
Schwusch, line 16: thought: For those services that regions are not applicable, it makes more sense to remove the `region` parameter as a whole, and "hard code" it on line 17.

Schwusch:
Great job on the PR! 
Just a few thoughts on the generated api for the affected services, but otherwise it's an impressive amount of work.
Thanks a lot for taking the time :)

generated/aws_iam_api/lib/iam-2010-05-08.dart
```diff
@@ -54,7 +54,7 @@ class IAM {
   final Map<String, _s.Shape> shapes;
 
   IAM({
-    @_s.required String region,
+    String region,
```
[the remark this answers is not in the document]
xvrh: I think it's a bit tricky. According to the config file, some region (like china) still have a custom end-point (https://github.com/aws/aws-sdk-js/blob/master/lib/region_config_data.json#L34)

generator/lib/builders/protocols/rest_xml_builder.dart
xvrh: I would be ok with that. It works for me. I'm just worried that some regions (like china or us-gov) may have a use case for a custom endpointUrl and region.

Schwusch: Hmm, right. How does one choose those regions in aws-sdk-js?

xvrh: I think there is a concept of a global configuration object where you can put the region (`AWS.config.region`).

xvrh: I just checked with aws-sdk-js:
For the IAM api, you can specify a region in the constructor, it will still use the global end point and the correct signing region.
But it also optional, the api works without a region specified.

```js
const iam = new AWS.IAM({
    apiVersion: '2010-05-08',
    region: 'eu-west-1', // works with and without
});
```

Schwusch: Oh well, I think that is good enough. I'd like @isoos to have a quick look as well, but it LGTM.

generator/lib/generate_command.dart
```diff
@@ -364,6 +368,20 @@ in the config file, from the downloaded models.''';
 
     return !(pr.stdout as String).contains('working tree clean');
   }
+
+  final _configDataFile = File('./apis/config/region_config_data.json');
+  Future<void> _generateConfigFiles() async {
+    final jsonContent =
```
isoos, line 374: I wish this would be first parsed to a typed data model, and then it would be run through building the config output.

shared_aws_api/lib/src/protocol/endpoint.dart
```diff
@@ -0,0 +1,138 @@
+import 'package:meta/meta.dart';
+import 'endpoint_config_data.dart' as config;
+import 'shared.dart';
+
+class Endpoint {
```
isoos, line 5: Could you please add some documentation what this does and what is it for? (The name endpoint is a bit vague).

shared_aws_api/lib/src/protocol/endpoint.dart
```diff
@@ -0,0 +1,138 @@
...
+      url.hashCode ^
+      signingRegion.hashCode ^
+      signatureVersion.hashCode;
+
+  @override
+  String toString() => 'Endpoint(service: $service, url: $url, '
+      'signingRegion: $signingRegion, signatureVersion: $signatureVersion)';
+}
+
+class RegionConfig {
```
isoos, line 94: Same documentation request here.

isoos (commented):
After a quick look at the code this looks in general okay.

generator/lib/generate_command.dart
xvrh: I added `model/region_config.dart` with a model `RegionConfigData` model.

shared_aws_api/lib/src/protocol/endpoint.dart
xvrh: I added some documentation. Feedback welcome.

shared_aws_api/lib/src/protocol/endpoint.dart
xvrh: I added some documentation. Feedback welcome.

Schwusch:
I think this looks fine, let's bring it in :)

Issue #245, opened 2020-10-21
XML List are not extracted correctly


xvrh:
For instance, the generated code for `ListRolePoliciesResponse` in the `IAM` API looks like:

```dart
  factory ListRolePoliciesResponse.fromXml(_s.XmlElement elem) {
    return ListRolePoliciesResponse(
      policyNames: _s
          .extractXmlChild(elem, 'PolicyNames')
          ?.let((elem) => _s.extractXmlStringListValues(elem, 'PolicyNames')),
    );
  }
```
https://github.com/agilord/aws_client/blob/7d4c87c83fdcdab81db1f0a0be57a805efb684c4/generated/aws_iam_api/lib/iam-2010-05-08.dart#L14691

However the returned response of AWS is:
```xml
<ListRolePoliciesResult>
  <PolicyNames>
    <member>CloudWatchPolicy</member>
    <member>DynamoDBPolicy</member>
  </PolicyNames>
</ListRolePoliciesResult>
```

So it should use `member` instead of `PolicyNames` in the `extractXmlStringListValues` call.
Or `extractXmlStringListValues` should not take the tag name and return the content of every children...

Schwusch:
The relevant code is located [here](https://github.com/agilord/aws_client/blob/master/generator/lib/builders/library_builder.dart#L425-L428).
It seems `memberElemName` should be conditional somehow. The answer to whether the name should be `member` or something else is probably in the API spec. I haven't looked in to it though.

Schwusch:
I made a PR with the `member` theory. Although the smartest move would be to check how the `aws-sdk-js` library does it, I haven't had the time to analyze it yet.

xvrh:
Thanks for working on this.
I think that this is what aws-sdk-js is doing: https://github.com/aws/aws-sdk-js/blob/master/lib/xml/node_parser.js#L100

To be able to be more confident for the fix of these kind of protocol mistake, I started to port the tests of the js sdk (test/fixtures/output).

You can see some of the progress here: https://github.com/xvrh/aws_client/commit/b18a39fdb5d95a4e5ba045557c9dc7b1cc25a882
It generates some deserialisation tests for the different protocols.

I plan to open a PR soon to discuss it further.

Schwusch:
> You can see some of the progress here: xvrh@b18a39f

[...]

PR #248, opened 2020-10-25
Xml lists are extracted incorrectly

Schwusch:
This PR is addressing #245, where list members seem to be extracted incorrectly.

### DIFF
--- a/generator/lib/builders/library_builder.dart
+++ b/generator/lib/builders/library_builder.dart
@@ -420,7 +420,7 @@ String _xmlExtractorFn(
     return '_s.extractXml${_uppercaseName(dartType)}Value($elemVar, \'$elemName\')${enumeration ? '?.to${parent.className}()' : ''}';
   } else if (type == 'list') {
     final memberShape = api.shapes[shapeRef.member.shape];
-    final memberElemName = shapeRef.member.locationName ?? elemName;
+    final memberElemName = shapeRef.member.locationName ?? 'member';
     String fn;
     if (memberShape.type.isBasicType()) {
       fn =
--- a/generated/aws_autoscaling_api/lib/autoscaling-2011-01-01.dart
+++ b/generated/aws_autoscaling_api/lib/autoscaling-2011-01-01.dart
@@ -4387,10 +4387,8 @@ class ActivitiesType {
   });
   factory ActivitiesType.fromXml(_s.XmlElement elem) {
     return ActivitiesType(
-      activities: _s.extractXmlChild(elem, 'Activities')?.let((elem) => elem
-          .findElements('Activities')
-          .map((c) => Activity.fromXml(c))
-          .toList()),
+      activities: _s.extractXmlChild(elem, 'Activities')?.let((elem) =>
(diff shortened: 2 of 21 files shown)


Schwusch:
@isoos Does this seem correct to you? Especially the SNS/SES changes. 
It seems weird that it would've been wrong for so long. On the other hand, the result of the faulty extraction seems to have been empty lists that hasn't raised enough eyebrows until now.

generator/lib/builders/library_builder.dart
```diff
@@ -420,7 +420,7 @@ String _xmlExtractorFn(
     return '_s.extractXml${_uppercaseName(dartType)}Value($elemVar, \'$elemName\')${enumeration ? '?.to${parent.className}()' : ''}';
   } else if (type == 'list') {
     final memberShape = api.shapes[shapeRef.member.shape];
-    final memberElemName = shapeRef.member.locationName ?? elemName;
+    final memberElemName = shapeRef.member.locationName ?? 'member';
```
xvrh, line 423: From aws-js-sdk: https://github.com/aws/aws-sdk-js/blob/master/lib/xml/node_parser.js#L100
It doesn't use `locationName` but `name` (that you don't have right now in this implementation).

isoos:
It would be nice to see an example response that is like this. Is there any documentation page that has it?

Schwusch:
> It would be nice to see an example response that is like this

@xvrh eloquently provided one in #245, but yeah, documentation seems sparse

generator/lib/builders/library_builder.dart
Schwusch: I think it's a little bit more complicated than that, I think `shape.member` is also a `shape`, and the `name` property is defined [here](https://github.com/aws/aws-sdk-js/blob/master/lib/model/shape.js#L28).
Not sure whether we should apply the same name hierarchy here though. I'm not sure what [`memberName`](https://github.com/aws/aws-sdk-js/blob/master/lib/model/shape.js#L29) is at that point of execution.

Schwusch:
I actually found a [PDF](https://docs.aws.amazon.com/IAM/latest/APIReference/iam-api.pdf) with an example response, page 280 :sweat_smile:

generator/lib/builders/library_builder.dart
xvrh: Oh ok. Pretty complicated yeah ;-)

xvrh:
Similarly, the same problem exists for extracting `Map`. 
It should default to `key` and `value` names.

You can search for `_s.extractXmlStringValue(c, 'null')` in the generated apis, you will find lots of incorrect extractions.

Schwusch:
I think I will mothball this change until #249 is in, so that we can have tests checking if we're doing the right thing.

xvrh:
@Schwusch yes we should wait. The tests reveals that the change is not correct in the case of "flattened" list.
Also, I have a fix for the Map extraction.

Schwusch:
> Also, I have a fix for the Map extraction.

We can close this PR if you have a more comprehensive fix, especially if it's based on test results :)

Schwusch:
Abandon in favor of #265

PR #249, opened 2020-10-26
Generate deserialisation test suite

xvrh:
Generates some deserialisation tests for the different protocols. Test cases are extracted from the aws-js-sdk project (test/fixtures/output/*).
This is to help find bugs like #248 

This is a first draft, see the discussion points bellow.

The tests exercises the entire pipeline: It creates a fake `Api` based on the test description and generates a Dart client.
It allows to verify that:
- The generator generates working code for all the shapes
- The generated code is able to deserialise the response and extract the correct data

As of now there are 21 tests that PASS and 34 that FAIL
From a quick look at the failures:
- Data are not extracted from headers correctly
- Blob data are not converted from base64
- Maps and list are not deserialised correctly
- Timestamp format are not respected everywhere

### Download the test suite
The .json tests files from the aws-sdk-js project are manually copied into the repository.

I actually started by implementing an automatic import in the download phase of the generator but I've backed-off because:
- There are some irreconcilable differences between JS and Dart and there are cases where this will be necessary to modify the test input/output. (ie: unknown enum)
- The test suite hasn't been modified in years

Maybe you have an opinion on this?

### Code generation
I added the testsuite generation along side the api generation phase (in `generate_command`).
Maybe you prefer something else:
- Create a custom command in bin/generate.dart
- Create an other entry point somewhere else

Also, should we automatically run the build_runner in the target project (shared_aws_api) after the generation?

Please let me know what you thing.

### How to manage failures?
If we commit this change as-it, the CI will fail every time until we fix all the issues.

We can try to generate a `skip: true` parameter on the tests that currently fail. When we fix them, we remove the skip parameter.


Schwusch:
A few points:
- Tests are failing because the protocol constructor argument `AwsClientCredentials` is null/missing. A simple 
```dart 
TestClass(credentials: AwsClientCredentials(accessKey:'', secretKey:''))
``` 
works when testing. 
- I think `build_runner` should be executed in `shared_aws_api` after tests has been generated. More automation is good.
- There is already tests for input, is it possible to apply this methodology to those as well to make the tests uniform? The tests at the moment are flawed, in that input classes are not used, thus some tests are failing(like blob base64 encoding and datetime parameters are not tested in the way we use them).

xvrh:
@Schwusch 
I have made some progress here. I looked at all the failures and I have a fix for most of them.
I'm still thinking what is the best way to push and review all the changes.

I plan to do the same thing for the input tests.

Schwusch:
> I'm still thinking what is the best way to push and review all the changes.

I think the split between generator and generated code changes into separate commits are fine.

> I plan to do the same thing for the input tests.

That's great, thank you :D

xvrh:
Hello @Schwusch ,
This change is ready to be reviewed. It generates the "input" test suite too.

This PR contains the minimal changes to build and run the suite without compilation error.
The test result is 101 FAILED and 78 PASS.

There are 3 commits: one that import the .json fixtures, one with the actual generator change and one with the generated files.

When it is merged, I'll open several other PR (one at a time) with some fixes (hopefully it will go down to 100% PASS).

generated/aws_events_api/test/ensure_build_test.dart
```diff
@@ -1,10 +1,10 @@
-import 'package:aws_events_api/eventbridge-2015-10-07.dart';
+import 'package:aws_events_api/events-2015-10-07.dart';
```
xvrh, line 1: This change slipped in. I don't know where it came from.

Schwusch: It comes from [this if-statement](https://github.com/agilord/aws_client/blob/master/generator/lib/generate_command.dart#L255-L267). It seems eventbridge/events share the same `serviceId` and `apiVersion`, which leaves the "latest built api version" to be random. I'm not sure what to do about that.

generator/lib/builders/test_suite_builder.dart
```diff
@@ -0,0 +1,234 @@
...
+          expectCode = "startsWith('$headerValue')";
+        }
+
+        code.writeln("  expect(request.headers['$header'], $expectCode);");
+      }
+    }
+    code.writeln("expect(pathAndQuery(request.url), '${request.uri}');");
+    if (request.host != null) {
+      //TODO: to support that, we need to be able to pass hostPrefix to instantiate
+      // the service
```
Schwusch, line 69: Might be good to create an issue for this, so we don't forget :)

Schwusch:
It seems dartanalyzer on Travis complains on [some missing parameters](https://travis-ci.com/github/agilord/aws_client/jobs/426882012). If that is fixed, the actual tests will run, which would be really interesting to see :D

generated/aws_events_api/test/ensure_build_test.dart
xvrh: I added a line to sort the files from "Directory().listSync()" in the generate command, maybe it will brings stability.

generator/lib/builders/test_suite_builder.dart
xvrh: Done #260

Schwusch:
Looks good! Let's bring it in

PR #250, opened 2020-10-26
Fix statusCode check

xvrh:
Some API returns 201 (created) and would then throw an exception due to this inversion.


isoos (approved):
ah, facepalm :)

PR #254, opened 2020-10-30
Use secure links in README

sidrao2006:
## Description

Use secure links in README to improve pub points

## Checklist

- [x] I am willing to follow-up on review comments in a timely manner.

### DIFF
--- a/generated/aws_sqs_api/README.md
+++ b/generated/aws_sqs_api/README.md
@@ -10,7 +10,7 @@ hosted queue for storing messages as they travel between applications or
 microservices. Amazon SQS moves data between distributed application
 components and helps you decouple these components.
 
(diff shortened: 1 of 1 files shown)


xvrh (changes requested):
This is a generated file that comes from the AWS API definition.
Precisely here: https://github.com/aws/aws-sdk-js/blob/master/apis/sqs-2012-11-05.normal.json#L1495

One possible solution is to change the generator to replace automatically the `http://` schemes to `https://` everywhere.

The code that generate the readme is here: https://github.com/agilord/aws_client/blob/9e67619d8717289d2b067b3b335dd059a59751ce/generator/lib/builders/readme_builder.dart#L12

sidrao2006:
Sorry, I'll open a new PR.

PR #257, opened 2020-10-30
Use secure links in README

sidrao2006:
Previously #254 

## Description

Use secure links in README to improve pub points

## Checklist

- [x] I am willing to follow-up on review comments in a timely manner.

### DIFF
--- a/generator/lib/builders/readme_builder.dart
+++ b/generator/lib/builders/readme_builder.dart
@@ -9,7 +9,7 @@ String buildReadmeMd(Api api) {
 **Warning: This is a generated library, some operations may not work.**
 
 *About the service:*
-${markdownText(api.documentation.replaceFirst('http', 'https'))}
+${markdownText(api.documentation.replaceFirst('http://', 'https://'))}
 
 ## Links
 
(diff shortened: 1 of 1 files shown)


sidrao2006:
cc @xvrh

generator/lib/builders/readme_builder.dart
```diff
@@ -9,7 +9,7 @@ String buildReadmeMd(Api api) {
 **Warning: This is a generated library, some operations may not work.**
 
 *About the service:*
-${markdownText(api.documentation)}
+${markdownText(api.documentation.replaceFirst('http', 'https'))}
```
xvrh, line 12: `replaceAll('http://', 'https://')`.
Else you will replace "https://" to "httpss://"

xvrh:
You can run the generator and see the result of your change with `dart bin/generate.dart generate --no-build` from the "generator" folder.

generator/lib/builders/readme_builder.dart
sidrao2006: Right, sorry

Schwusch:
I believe many links are probably fixed in newer versions of aws-sdk-js, the version we are using is [6 months old](https://www.npmjs.com/package/aws-sdk/v/2.659.0) at this time of writing. Perhaps we should look at bumping that version some time, but I am afraid it will bring a lot of work to catch up.

Schwusch:
This is solved in the generator nowadays

PR #259, opened 2020-11-02
Optimise build_runner phase in package generator

### LINKED ISSUE #258: build_runner on +200 packages is very slow
Running `pub get` and `pub run build_runner build` on the +200 packages takes a lot of time (>30 minutes on my mac?).

Raw idea: 
- Generate all the API's in a single development package where each API has its own folder.
```
all_apis
  lib
    aws_accessanalyzer_api
      accessanalyzer-2019-11-01.dart
```

- Run `build_runner` on this `all_apis` package where it will generate the `.g.dart` files.
- Have an other command that copy the content of each folder and generate the corresponding package.

The `all_apis` package is not necessary versioned.

WDYT?

xvrh:
Fixes #258 

This PR refactors the generate command so the `build_runner` phase is only run once. 
It generates all the apis in a temporary `all_apis` package, runs `build_runner` on it and then copies the generated file in their final destination.

On my machine, I can run the `generate` command in less than 2 minutes from a clean repository (and 30 seconds on the next run).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ doc/api/
 
 # Generated directories
 **/apis/
+/all_apis/lib/
--- a/all_apis/pubspec.yaml
+++ b/all_apis/pubspec.yaml
@@ -1,0 +1,16 @@
+name: all_apis
+publish_to: none
+
+environment:
+  sdk: '>=2.7.0 <3.0.0'
+
+dependencies:
+  shared_aws_api:
+    path: ../shared_aws_api
+
+dev_dependencies:
+  build_runner:
+  json_serializable: ^3.3.0
+  test: ^1.14.2
+
+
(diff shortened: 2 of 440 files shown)


Schwusch:
This is extremely positive! This makes so much other stuff possible...
I see you've removed `build_runner` from the generated packages, I don't think `json_serializable` and `build_verify` can be used without it. Might as well remove those two as well :)

I also have a suggestion for our way of working: split the work into two commits, one with changes to shared/generator, and one with the result of running `generate` after those changes. It would make the reviewing much easier.

xvrh:
> Might as well remove those two as well

Done

> split the work into two commits

Done. The first commit contains the manual changes

Schwusch:
I just realized `dart bin/generate.dart test` will probably fail because of the missing dependencies in the packages. Other than that I think it looks good.
Do you want to fix those here or should we fix it later?

Schwusch (approved):

xvrh:
Thank you for the test fix.

Schwusch:
After fixing and running the tests, I'm starting to question the tests in the generated packages :)
The point is to see that they compile, so that there are no obvious syntax or type errors. That could probably be done in the `all_apis` package as well...

PR #265, opened 2020-11-05
Fix Xml list and map extraction

### LINKED ISSUE #245: XML List are not extracted correctly

xvrh:
Fixes #245 

Fix XML extraction of List elements and Map entries.

First commit is the actual generate code change, second commit the impact on the generated code.

10 more tests are now succeeding.

### DIFF
--- a/generated/aws_autoscaling_api/lib/autoscaling-2011-01-01.dart
+++ b/generated/aws_autoscaling_api/lib/autoscaling-2011-01-01.dart
@@ -4394,10 +4394,8 @@ class ActivitiesType {
   });
   factory ActivitiesType.fromXml(_s.XmlElement elem) {
     return ActivitiesType(
-      activities: _s.extractXmlChild(elem, 'Activities')?.let((elem) => elem
-          .findElements('Activities')
-          .map((c) => Activity.fromXml(c))
-          .toList()),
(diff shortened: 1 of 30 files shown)


Schwusch (approved):
Looks good, you're killing it with those tests :)

PR #268, opened 2020-11-09
Read and write Xml Attribute

xvrh:
Respect the `Member.xmlAttribute`

Fixes 2 tests but have small impact on the actual API's.

First commit is the actual code change, second is the re-generated apis.


Schwusch:
Nice, LGTM 🤘

PR #271, opened 2020-11-12
Use input shape for json protocol

xvrh:
This PR refactors the json protocols to generate the "Input shape" and use it for the serialisation of the request body.

This is so we don't have to handle again the DateTime, enum, blob fields. We have it for free with json_serializable.

First commit is the actual code change, second is the re-generation of the apis.

Fixes 4 tests in the testsuite.


Schwusch:
Haha soon we'll need a new strategy for code reviewing, seems like we're hitting the limits of Github in a lot of aspects.
I can't even list the changed files...
I'll do a CR locally on my machine then :smile: 
image

Schwusch:
One thought that strikes me when seeing changes this large, I'd like to know that the resulting code actually compiles.
One way I've tried to verify this is through instantiation tests in each package, which is very slow to run , but can this be verified when creating `all_apis` package perhaps? There has been occasions where we've published broken packages before...

xvrh:
> I can't even list the changed files

Yes it is not pleasant to review inside Github. Although only the first commit is the manual code change (which is much smaller).

> I'd like to know that the resulting code actually compiles

Good idea, I'll send a new pull request to setup an analyse/test of the all_apis package on CI.

xvrh:
After thinking about it a bit, I'm not sure if this PR is the right path forward.

Maybe the opposite direction: have the entire serialisation/deserialisation generated in the Operation code is more sensible.

I'll investigate and possibly not merge this.

xvrh:
Closing in favour of an alternate approach taken in #280

PR #275, opened 2020-11-13
Use signingName to sign request

### LINKED ISSUE #272: Use api.metadata.signingName if not null
`signingName` should have priority over `endpointPrefix` to compute the signature.

There are about 40 apis impacted

In aws-js-sdk: https://github.com/aws/aws-sdk-js/blob/4fd99d087f245f23b164219fba62156af060770e/lib/service.js#L462

xvrh:
Use the field `signingName` from api.metadata to sign the request (fallback to `endpointPrefix` if not present).

First commit is the code change to review and the second is the re-generated apis.

Fixes #272

### DIFF
--- a/generated/aws_accessanalyzer_api/lib/accessanalyzer-2019-11-01.dart
+++ b/generated/aws_accessanalyzer_api/lib/accessanalyzer-2019-11-01.dart
@@ -44,7 +44,10 @@ class AccessAnalyzer {
     String endpointUrl,
   }) : _protocol = _s.RestJsonProtocol(
           client: client,
-          service: 'access-analyzer',
+          service: _s.ServiceMetadata(
+            endpointPrefix: 'access-analyzer',
+            signingName: 'access-analyzer',
+          ),
           region: region,
           credentials: credentials,
           endpointUrl: endpointUrl,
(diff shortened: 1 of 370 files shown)


Schwusch:
I haven't run in to an issue with this before, and there doesn't seem to be any tests that touch on it.
Which use-cases does this solve? I don't doubt it fixes something, I'm just curious.

xvrh:
Hi Jonathan,
Thank you for looking into this actually, it's good to have a second pair of eyes to help.

I was trying to use the IoT-Data api and went into this error: `403 ForbiddenException: Credential should be scoped to correct service`

Before this PR, the protocol class uses the `api.metadata.endpointPrefix` for 2 things: fill the `{service}` part of the endpointUrl (`{service}.{region}.amazonaws.com`) and sign the request (`signAws4HmacSha256`).

When looking into the `aws-js-sdk`, I discovered that there is the `api.metadata.signingName` that is preferred to sign the request. (https://github.com/aws/aws-sdk-js/blob/4fd99d087f245f23b164219fba62156af060770e/lib/service.js#L462)

My deduction is that the 40 services that declares a `signingName` different than `endpointPrefix` would currently fails. It would be excellent if you can test one or too on your side to validate the good effect of this PR.

Schwusch:
I see! Well the services I've used will have the same behaviour after this change, and the rest seems correct, derived from aws-sdk-js behaviour and your empirical tests.
LGTM and nice find!

Schwusch (approved):

PR #277, opened 2020-11-16
Use enum in List and Map

xvrh:
Use an enum for members of type `List` and `Map`.

There are still a few bugs (ie. extraction from headers) but I would like to have this first part landed first before into further details.


Schwusch:
I noticed 2 more tests failing on this PR, but I suppose you have a longer term plan :)
LGTM

Schwusch (approved):

xvrh:
Thanks to keep the counting @Schwusch :-)
This is an intermediate step, I plan to fix more tests soon.

PR #278, opened 2020-11-17
[Rest-Json] Support blob payload and non-string headers

xvrh:
This change is to fix 2 things:

### Blob payload in Rest-Json protocol

Some rest-json endpoints have raw bytes in the http response body and thus should not be json decoded.

I expose a `Future<http.Response> sendRaw()` method in the protocol so the generated operation can directly read the bytes of the response.

The Rest-xml protocol has the same problem and will need this change too in the future.

### Headers data in Rest-Json

Before this change, the `response.headers` are merged with the decoded json body. The new map is passed to the fromJson() factory (generated by json_serializable).

This is actually wrong. `response.headers` is a `Map<String, String>`. The values are not encoded in json and only String values would  work (other types throws a cast error).
I couldn't find a simple workaround. I went to generate the full extraction of json keys and headers inside the operation method. This also requires to generate the enum helpers for this cases.

Let me know if you have any concern.

### DIFF
--- a/generated/aws_apigateway_api/lib/apigateway-2015-07-09.dart
+++ b/generated/aws_apigateway_api/lib/apigateway-2015-07-09.dart
@@ -2375,7 +2375,7 @@ class APIGateway {
     _query = '?${[
       if (parameters != null) _s.toQueryParam(null, parameters),
     ].where((e) => e != null).join('&')}';
-    final response = await _protocol.send(
+    final response = await _protocol.sendRaw(
       payload: null,
       headers: headers,
       method: 'GET',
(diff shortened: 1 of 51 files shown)


Schwusch:
> This is actually wrong. response.headers is a Map<String, String>. The values are not encoded in json and only String values would work (other types throws a cast error).

In which situation would there be other types than Strings in the headers? Have you encountered a situation where this happens?

xvrh:
There are a few cases of integer and DateTime. You can find them by searching for `extractHeaderIntValue` in the diff of this PR.

The values are stored as String in the `Map` but the Json extractor contains code like: `json['Content-Length'] as int`.

Schwusch (approved):

xvrh:
Thank you Jonathan.

PR #279, opened 2020-11-17
Use Uint8List for blob in query protocol

xvrh:
I was wondering why `blob` are generated as `String` in the query protocol instead of `Uint8List`?

So I decided to ask with this pull request :-)


Schwusch:
I don't know actually.
@isoos might know?

Schwusch (approved):

PR #280, opened 2020-11-22
Fixes for Json root input encoding

xvrh:
Fixes the JSON encoding of the root input shape (JSON and Rest-JSON protocols):
- Respect "locationName"
- Encode complex `Map` and `List` (ie. `List<Enum>` needs to be converted in `List<String>`)
- Encode `DateTime`

Fixes 11 tests from the test suite

As always, first commit is the actual generator change and second commit is the generated apis.

### DIFF
--- a/generator/lib/builders/protocols/rest_json_builder.dart
+++ b/generator/lib/builders/protocols/rest_json_builder.dart
@@ -2,7 +2,7 @@ import 'package:aws_client.generator/builders/protocols/service_builder.dart';
 import 'package:aws_client.generator/model/api.dart';
 import 'package:aws_client.generator/model/operation.dart';
 import 'package:aws_client.generator/model/shape.dart';
-import '../../model/dart_type.dart';
+import '../builder_utils.dart';
 import '../library_builder.dart';
 
 class RestJsonServiceBuilder extends ServiceBuilder {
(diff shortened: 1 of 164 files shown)


generator/lib/builders/protocols/rest_json_builder.dart
```diff
@@ -151,3 +146,35 @@ String extractJsonCode(Shape shape, String variable, {Member member}) {
...
+    return '${timestampFormat}ToJson($variable)';
+  } else if (shape.type == 'blob') {
+    if (maybeNull) {
+      return '$variable?.let(base64Encode)';
+    } else {
+      return 'base64Encode($variable)';
+    }
+  }
+  return '$variable';
+}
```
Schwusch, line 180: Perhaps this should be broken out into some kind of `builder_utils` file, it looks a little weird when `json_builder` has `import 'rest_json_builder.dart';`

Schwusch:
LGTM, just a nit comment but otherwise, really good job!

generator/lib/builders/protocols/rest_json_builder.dart
xvrh: Done. Thanks for the idea.

Schwusch (approved):

PR #281, opened 2020-11-27
Blob response in Xml protocol and more

xvrh:
- Blob response in Xml protocol
- Raw string and blob input in Rest-Json
- Encode DateTime (and respect timestamp format), Enums, List and map in headers and Query
- Serialize Map in Xml protocol

Fixes 15 tests in the test suite!

One of the next step will be to consolidate the code generator for the headers and query, and share it between all protocols.


Schwusch:
I was actually working on some of these fixes the other night, but you did it much better than I ever could :)
LGTM and good job!

Schwusch (approved):

xvrh:
Thanks Jonathan.

> I was actually working on some of these fixes the other night

Oh, sorry for the duplicate work. To prevent that in the future, I will create some github issues for the remaining failures and announce what I plan to work on.

PR #282, opened 2020-11-27
Add request.query to the signature

xvrh:
Some endpoint with query in the requestUri fails with:
`The request signature we calculated does not match the signature you provided. Check your key and signing method`


Schwusch:
Nice catch, I am not able to test that this fix works, but I trust you have tested it :)

PR #287, opened 2020-12-04
Fixes query parameters

### LINKED ISSUE #286: [Test suite failures] Path and query parameters
About 6 tests related to how we encode query parameters and path parameters still fail.

- Encode each query parameter (specially those encoded with a custom timestamp format)
- "Greedy keys"
- Recognise enums in "path parameters"
- List in query parameters (`?List=1&List=2` => it requires to use a custom class instead of a `Map`)

xvrh:
Fixes #286 

- Correctly encode List and Map in query parameters
- Encode enum in path
- Respect greedy path component (`{+path}`)

It fixes 9 failures in the test suite.

### DIFF
--- a/generated/aws_accessanalyzer_api/lib/accessanalyzer-2019-11-01.dart
+++ b/generated/aws_accessanalyzer_api/lib/accessanalyzer-2019-11-01.dart
@@ -183,8 +183,7 @@ class AccessAnalyzer {
     await _protocol.send(
       payload: $payload,
       method: 'PUT',
-      requestUri:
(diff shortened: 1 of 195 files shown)


Schwusch:
I pulled in #282 and it seems it caused a conflict. 
The changes looks good to me, really good job @xvrh :D You're a hero, doing this work.

Schwusch (approved):

xvrh:
Thanks Jonathan. Less than 10 failures remaining... Almost there

PR #289, opened 2020-12-07
Add accept header in Rest-Json protocol

### LINKED ISSUE #288: DateTime from server doesn't always respect protocol format
I notice some `rest-json` protocol returns some DateTime encoded as iso8610. 
While this library try to interpret it as unixTimestamp 
https://github.com/agilord/aws_client/blob/e68105fae4527f0fe44487fcc18de48976b55648/generator/lib/builders/library_builder.dart#L254

I see that aws-js-sdk has a parse function that try to support every format:
https://github.com/aws/aws-sdk-js/blob/7d4dd548f74f739dbe0df1c4b9cd88454ce5d131/lib/util.js#L330

xvrh:
The `rest-json` protocol respond differently if the `Accept` header is present or not.

For instance, with the `apigateway-2015-07-09`
```dart
  final apigateway = APIGateway(region: 'eu-west-1');
  final api = await apigateway.getRestApi(restApiId: 'xx');
```

Without Accept: application/json
```json
{
    "_links": "....",
    "apiKeySource": "HEADER",
    "createdDate": "2019-12-05T12:47:19Z",
    "disableExecuteApiEndpoint": false,
    "endpointConfiguration": {
      "vpcEndpointIds": null,
      "types": "EDGE",
      "ipv6": false
    },
    "id": "xxx",
    "..."
  }
```

With Accept: application/json
```json
  {
    "apiKeySource": "HEADER",
    "createdDate": 1575550039,
    "disableExecuteApiEndpoint": false,
    "endpointConfiguration": {
      "ipv6": false,
      "types": [
        "EDGE"
      ]
    },
    "id": "xxx",
    "..."
  }
```

- The date is in the correct format (fixes #288)
- The "types" entry is now (correctly) a json list (previously would fail with a CastError).

### DIFF
--- a/shared_aws_api/lib/src/protocol/rest-json.dart
+++ b/shared_aws_api/lib/src/protocol/rest-json.dart
@@ -60,6 +60,7 @@ class RestJsonProtocol {
         rq.body = json.encode(payload);
       }
     }
+    rq.headers['accept'] = 'application/json';
     if (headers != null) {
       rq.headers.addAll(headers);
     }


Schwusch:
haha I simultaneously did #290, but this is good also, since it could probably introduce other bugs otherwise :+1:

Schwusch (approved):

PR #290, opened 2020-12-07
Parse dates in a generic manner

### LINKED ISSUE #288: DateTime from server doesn't always respect protocol format

Schwusch:
This fixes #288 and also seems to fix 2 tests somehow.


xvrh (approved):
Great. It's good to follow what aws-sdk-js is doing.

Schwusch:
Weird, it didn't fix the tests when running on Jenkins...
On my machine it were only 8 tests failing. I'll have to investigate why, but I think it's the tests that fail type casting.

Schwusch:
I see, the tests are running on Travis with the `--run-skipped` argument, which explains why it has more failures. Perhaps we should remove that argument?

Issue #285, opened 2020-11-28
[Test suite failures] Output complex Map values


xvrh:
One output test is still failing because we fail to decode `Map<String, DateTime>` with a custom timestamp format.

One of the solution is to create a custom `JsonConverter` for each timestampFormat and apply it to the properties that need it.

Schwusch:
I just did some investigation around this, and it seems no services has maps of DateTimes, only that test.

xvrh:
There are 2 json apis with `List<DateTime>` (QuickSight and ComputeOptimizer) that would fail currently like the Map.
The solution with the JsonConverter would fix both map and list. I think we should keep the test.

Schwusch:
Should we create a test that actually tests the list case then?

Schwusch:
I added a case for the datetime lists and a fix for that case in #291

PR #291, opened 2020-12-07
Don't run skipped tests in shared_aws_api, and convert complex map test to list

Schwusch:
This "solves" #285


.travis.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       env: PKGS="shared_aws_api"
       script: ./tool/travis.sh dartfmt dartanalyzer
     - stage: unit_test
-      name: "SDK: stable; PKG: document_client; TASKS: `pub run test -j 1 --run-skipped`"
+      name: "SDK: stable; PKG: document_client; TASKS: `pub run test -j 1`"
```
xvrh, line 31: Good! I was wondering since the beginning why the numbers where not the same locally and in travis.

shared_aws_api/test/fixtures/output/rest-json.json
```diff
@@ -402,6 +402,7 @@
   },
   {
     "description": "Complex Map Values",
+    "skip": "No api actually uses this, only this test",
```
xvrh, line 405: I think we should not skip this test, because the List<DateTime> have the same problem and do exist in the apis.

Schwusch: I think we should make a proper test with the list case :)

xvrh: Yes we should!
My idea was that a solution for Map<DateTime> would most probably cover the List case but it's better to have a real test.

Schwusch:
@xvrh I updated the test, WDYT? I know this makes the fixtures diverge from the source, but from my understanding, these tests aren't updated very often?

Schwusch:
I added a fix for the test case as well, I hope it fits in this PR

Schwusch:
I am unable to format the code according to Dart stable channel, since I am on the dev channel. @xvrh what channel are you using, should we use dev channel on Travis since we probably need to start thinking of NNBD?

xvrh:
@Schwusch I mainly use version `2.10.4`. But I also have other versions available as shell aliases. So feel free to use whatever work for you.

shared_aws_api/lib/src/protocol/shared.dart
```diff
@@ -49,6 +49,33 @@ DateTime timeStampFromJson(dynamic date) {
   throw ArgumentError.value(date, 'date', 'Unknown date type, can not convert');
 }
 
+abstract class DateTimeListConverter
```
xvrh, line 52: I think that there is an alternative implementation to consider:

If you define a 
```dart
class DateTimeConverter implements JsonConverter<DateTime, dynamic>
```
and apply it to a List or Map field, `json_serializable` will recognize it correctly.

Furthermore, we can apply a `@XXXDateTimeConverter()` directly on the class member.

So we can generate:

```dart
class ApiResult {
  @RfcDateTimeConverter()
  List<DateTime> listDates;

  @IsoDateTimeConverter()
  Map<String, DateTime> mapDates;

  @UnixDateTimeConverter()  // Optional: replace the current way with fromJson/toJson
  DateTime simpleDate;
}
```

In the generator, the condition `if (member.dartType == 'DateTime')` becomes  (more or less)
```dart
if (member.dartType == 'DateTime' ||
              (member.shapeClass.type == 'list' &&
                  member.shapeClass.member.shapeClass.type == 'timestamp') ||
              (member.shapeClass.type == 'map' &&
                  member.shapeClass.value.shapeClass.type == 'timestamp'))
```

Schwusch: Wow, I didn't know json_serializable picked up converters on "partial" types, but it totally worked!
Thanks, I updated the code accordingly

xvrh (approved):
Nice!

shared_aws_api/lib/src/protocol/shared.dart
xvrh: 👍 
FYI an other case where we can use this feature of json_serializable: `Uint8ListListConverter` (https://github.com/agilord/aws_client/blob/master/shared_aws_api/lib/src/protocol/shared.dart#L72)

PR #294, opened 2020-12-20
Idempotency token autofill

### LINKED ISSUE #283: [Test suite failures] Idempotency token
4 failures comes from "Input / Idempotency token auto fill" tests.

From what I understand, some members are marked with `idempotencyToken: true`. If not specified by the users, those members should be autofilled by the library with a generated uuid.

The test mocks the uuid generator to always use `00000000-0000-4000-8000-000000000000`

Schwusch:
Fixes #283

### DIFF
--- a/generator/lib/builders/protocols/json_builder.dart
+++ b/generator/lib/builders/protocols/json_builder.dart
@@ -29,14 +29,16 @@ class JsonServiceBuilder extends ServiceBuilder {
 
     final payloadMembers = inputClass?.members?.map((member) {
       final buffer = StringBuffer();
-      if (!member.isRequired) {
+      if (!member.isRequired && !member.idempotencyToken) {
         buffer.writeln('if (${member.fieldName} != null)');
       }
       final encodeCode = encodeJsonCode(member.shapeClass, member.fieldName,
           member: member, maybeNull: member.isRequired);
       final location =
           member.locationName ?? member.shapeClass.locationName ?? member.name;
-      buffer.writeln("'$location': $encodeCode,");
+      final idempotency =
+          member.idempotencyToken ? '?? _s.generateIdempotencyToken()' : '';
+      buffer.writeln("'$location': $encodeCode$idempotency,");
       return '$buffer';
     })?.join();
     var payload = '';
--- a/generator/lib/builders/protocols/rest_json_builder.dart
+++ b/generator/lib/builders/protocols/rest_json_builder.dart
@@ -40,7 +40,7 @@ class RestJsonServiceBuilder extends ServiceBuilder {
         if (inputShape.hasBodyMembers) {
           buf.writeln('final \$payload = <String, dynamic>{');
           for (var member in inputShape.members.where((m) => m.isBody)) {
-            if (!member.isRequired) {
+            if (!member.isRequired && !member.idempotencyToken) {
               buf.writeln('if (${member.fieldName} != null)');
             }
             final encodeCode = encodeJsonCode(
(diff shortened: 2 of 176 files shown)


generator/lib/builders/protocols/json_builder.dart
```diff
@@ -14,8 +14,10 @@ class JsonServiceBuilder extends ServiceBuilder {
     final regionRequired = api.isGlobalService ? '' : '@_s.required';
     return '''
   final _s.JsonProtocol _protocol;
-  ${api.metadata.className}({$regionRequired String region, _s.AwsClientCredentials credentials, _s.Client client, String endpointUrl,})
-  : _protocol = _s.JsonProtocol(client: client, service: ${buildServiceMetadata(api)}, region: region, credentials: credentials, endpointUrl: endpointUrl,);
+  final String Function() _idempotencyTokenGenerator;
+  ${api.metadata.className}({$regionRequired String region, _s.AwsClientCredentials credentials, _s.Client client, String endpointUrl, String Function() idempotencyTokenGenerator,})
```
xvrh, line 18: Could we only add the `idempotencyTokenGenerator` parameter, only if there is a member with `idempotency: true`?

Personal opinion: I find it a bit "too much" to add this parameter. It will not be used very often (probably only in our tests).
I like the classical Flutter's approach to allow that kind of behaviour to be overridden in the tests: simply add a global variable to allow the override.
```dart
String Function() idempotencyGeneratorOverride;
String generateIdempotencyToken() => (idempotencyGeneratorOverride ?? _uuidV4)();
```

I find it is a bit less intrusive. At the end, this is definitely your call :-)

generator/lib/builders/protocols/rest_json_builder.dart
```diff
@@ -40,7 +42,7 @@ class RestJsonServiceBuilder extends ServiceBuilder {
         if (inputShape.hasBodyMembers) {
           buf.writeln('final \$payload = <String, dynamic>{');
           for (var member in inputShape.members.where((m) => m.isBody)) {
-            if (!member.isRequired) {
+            if (!member.isRequired && !member.idempotencyToken) {
```
xvrh, line 45: Is is possible to have a member with idempotencyToken in a Shape that is not the "root input shape"?

```dart
class Struct {
  String token; // Is is possible?
}
class Api {
  Future operation(Struct struct) {
    send({
      'struct': struct.toJson()
    });
  }
}
```

If it doesn't happen in practice in the api: then it is not necessary to support it BUT it may be safer to still detect the case and throw an UnimplementedError in case it happen in a future update.

xvrh (approved):
I added 2 small things that concern me. Feel free to take the final decision.

generator/lib/builders/protocols/json_builder.dart
Schwusch: Where would the global variable live, in the shared library? 
If multiple services are used, only one instance of the variable would exist then right?

generator/lib/builders/protocols/rest_json_builder.dart
Schwusch: Well everything is possible, but I haven't seen any instance where an idempotency token isn't in the root shape.
One thing that seems to correlate, but I am unable to verify it strictly, is that the operation always is marked `idempotent: true` if the root shape has `idempotencyToken: true` on one of it's members. I can investigate that further.

generator/lib/builders/protocols/json_builder.dart
xvrh: I was imagining it in the shared library as a global variable shared between every services.
The Flutter package does that a lot with variables named `debugXXXOverride` mainly to allow to change some behaviour in tests.

The user always have the option to directly pass the "token" value.
Do you think there is a use-case for the developer to inject the generator function for each service?

generator/lib/builders/protocols/rest_json_builder.dart
xvrh: OK thank, let's leave it like that. It's probably not important.

generator/lib/builders/protocols/json_builder.dart
Schwusch: > Do you think there is a use-case for the developer to inject the generator function for each service?

No you're right, I'll put it in the shared library.

PR #296, opened 2021-01-01
Fix json value trait tests

### LINKED ISSUE #284: [Test suite failures] Json trait
2 tests related to "Json trait" are failing.

Some members are marked with `jsonvalue: true`. Those members should have a `Object` dart type and be automatically json encoded/decoded by the library.
For input in header, the value should also be base64 encoded.

Schwusch:
Partially fixes #284. There is still one failing test with lists of `jsonValue` objects.
I am doing this PR mostly to get the discussion going on whether it's a reasonable approach.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -2,6 +2,7 @@
 include: package:pedantic/analysis_options.yaml
 
 analyzer:
+  exclude: [generated/**]
   errors:
     unused_element: error
     unused_import: error
--- a/generator/lib/builders/builder_utils.dart
+++ b/generator/lib/builders/builder_utils.dart
@@ -5,7 +5,10 @@ import '../model/shape.dart';
 import '../utils/string_utils.dart';
 
 String extractJsonCode(Shape shape, String variable, {Member member}) {
-  if (member?.jsonvalue == true) {
+  if (member?.jsonvalue == true || shape.member?.jsonvalue == true) {
+    if (shape.type == 'list') {
+      return '$variable == null ? null : ($variable as List).map((v) => jsonDecode(v as String)).toList().cast<Object>()';
+    }
     return '$variable == null ? null : jsonDecode($variable as String)';
   } else if (shape.type == 'map') {
     return '($variable as Map<String, dynamic>)?.map((k, e) => MapEntry(k, ${extractJsonCode(shape.value.shapeClass, 'e')}))';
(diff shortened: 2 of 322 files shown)


Schwusch:
Darn. I didn't see the analyzer warnings locally, and the tests actually ran as supposed to :man_shrugging: 
Weird that the tests does not fail locally due to type faults.

analysis_options.yaml
```diff
@@ -2,6 +2,7 @@
 include: package:pedantic/analysis_options.yaml
 
 analyzer:
+  exclude: [generated/**]
```
xvrh, line 5: What is the idea with this exclusion?

generator/lib/builders/builder_utils.dart
```diff
@@ -5,7 +5,9 @@ import '../model/shape.dart';
 import '../utils/string_utils.dart';
 
 String extractJsonCode(Shape shape, String variable, {Member member}) {
-  if (shape.type == 'map') {
+  if (member?.jsonvalue == true) {
```
xvrh, line 8: I think we should also look at the `Descriptor` class that has a `jsonvalue` property

xvrh (commented):
Thanks for doing this Jonathan. 
I started something on my side last week.
I push it now so you can have a look if you want: https://github.com/xvrh/aws_client/commit/199acccefc9631a13d952d1a1c1a4c7385f5a87c.

analysis_options.yaml
Schwusch: While developing, it isn't very useful to let the analysis engine plow through those packages.
We already do analysis on the `all_apis` package, and when excluding the `generated` folder, actually useful warnings and errors is showed when opening the project at its root.

If I understood it correctly, we can have an `analysis_options.yaml` in the `generated` folder too, which will be read when opening that folder in an editor or running the dart analyzer on that folder or one of its subfolders.

For me, concretely, the difference was going from ~15 000 warnings and errors to only the 8 that I cared about in this PR.
The speedup was also incredible, and the Dart plugin responsiveness when developing in the generator or shared package was a 100-fold better when excluding the `generated` folder.

xvrh: Thanks for the explanation!
I forgot about this problem because I already excluded this folder with an IntelliJ feature.

generator/lib/builders/builder_utils.dart
Schwusch: Right, thanks for the pointer. Managed to fix the last test case thanks to that :+1:

xvrh (approved):
Nice to see the CI to pass entirely!!

Schwusch:
Yes that's a lovely green light ♥️

PR #306, opened 2021-03-25
propagateTags for a Service in ECS can also have a value of NONE

thelmuxkriovar:
I've also submitted an issue in the js sdk this generates the definitions from https://github.com/aws/aws-sdk-js/issues/3677


Schwusch:
When the JS SDK definitions are updated, please open an issue to regenerate the service with the JS SDK version that contains the fix.
Manual updates will be overwritten when generating the services next time, so it's not recommended.

Issue #293, opened 2020-12-20
Proposal: drop json_serializable dependency


xvrh:
It seems we have all the code necessary to [encode](https://github.com/agilord/aws_client/blob/master/generator/lib/builders/builder_utils.dart#L26) and [decode](https://github.com/agilord/aws_client/blob/master/generator/lib/builders/builder_utils.dart#L7) `Shape` in json. We use this code to serialize/deserialize the root Input and Output shape of an Operation.

I did a quick experiment and was able to drop the json_serializable generator and replace the `fromJson` and `toJson` method with this implementation and all the tests still work. It is a 10 LoC change.

The advantages that I see:
- Faster code generation (no build_runner to start)
- Consolidate to a single implementation of json encoding
- Less code: all the JsonConverters can be deleted

This is just an idea. Let me know what you think.
 I may be missing something important to consider?

Schwusch:
I'm not sure I fully understand. Could you give an example of what would be migrated from `json_serializable` to `encodeJsonCode`/`extractJsonCode`?

xvrh:
Hi Jonathan,
Sorry for not being clear, I'll try to clarify :-)

This is more or less what we generate now in rest-json api:
```dart
class Api {
  Future<OutputShape> operation(Struct param1, SomeEnum param2) async {
    var response = await _protocol.send(body: {
      'param1': param1.toJson(),
      'param2': param2.toValue(),
    });
    return OutputShape({
      field: response.headers['param1'],
      field2: Struct.fromJson(response.body),
    });
  }
}

class Struct {
  String field1;
  SomeEnum field2;
  
  Struct();
  
  // Current: Delegate fromJson/toJson to json_serializable
  factory Struct.fromJson(json) => _$StructFromJson(json);
  Map toJson() => _$StructToJson(this);
}
```

### Proposal
Remove the json_serializable step and generate directly fromJson/toJson for the Structure shapes.
```dart
class Struct {
  String field1;
  SomeEnum field2;

  Struct();

  // Proposal: generate fromJson/toJson methods directly
  factory Struct.fromJson(json) {
    return Struct(
      field1: json['field1'],
      field2: json['field1'].toSomeEnum(),
    );
  }

  Map toJson() => {
        'field1': field1,
        'field2': field2.toValue(),
      };
}
```

### Reasoning
We already have all the code needed to generate those methods because we have to generate it in the operation method.

### Experiment
In `library_builder.dart`, replace:
```dart
[...]

Issue #292, opened 2020-12-09
Migrate to null safety


Schwusch:
We will have to support null safety at some point, and this ticket is for tracking that work.

Schwusch:
At the time of writing, `shared_aws_api` is listed as the 7:th most blocking package blocking null safety migration:
https://github.com/dart-lang/sdk/wiki/Null-safety-migration-status

Schwusch:
As it stands, we have 3 dependencies that are not migrated:
- package:crypto - https://github.com/dart-lang/crypto/issues/105
- package:http - https://github.com/dart-lang/http/issues/501
- package:build_runner - https://github.com/dart-lang/build/issues/2920

It seems `package:http` is close to a migrated release, and `package:crypto` seems to be migrated but no release.
I don't think `package:build_runner` needs to be migrated though, since it only runs when generating code?

IchordeDionysos:
@Schwusch looks like you are now unblocked and should be able to migrate to null safety :) 
Also you are now the 2nd most blocking package after `dio`

pshichenko:
Any way I can help with this?

Schwusch:
@psichenko There is a branch, `null_safety`, that you are welcome to fork and work on.
I currently have migrated the shared package, and there is some attempts to generate null safe code in the generator.
First step is to try to run the generator and observe the changes.
Feel free to ask questions on Gitter at dart-lang/server if you need quick answers.

xvrh:
Hi Jonathan, if you are not working on it right now, I may give it a try starting from you branch.

Schwusch:
I guess we can close this now as the null safe services has been published as 1.0.0 in #324 
Better late than never :sloth: :runner: 
Thanks everyone who helped, @xvrh in particular :heart: 
![itsdone](https://user-images.githubusercontent.com/8165610/146456272-3dce264f-e3a0-4548-a672-b26f1bf1dbf2.gif)

PR #308, opened 2021-04-14
Migrate to null safety

xvrh:
This is my attempt to migrate the generated apis to null-safety

This was not easy!

I had to make a lot of changes and some changes may be controversial and need to be adjusted based on feedbacks.

For instance, I choose to make Map and List elements non-nullable. I think it gives a more user-friendly API. But it also mean that we can no longer send and receive null elements in list and map.
I would like to learn the API that are affected by this limitation and try to find a way to "opt-in" the nullability rather than make all APIs upfront nullable.
(I think googleapis package has the same question: https://github.com/google/googleapis.dart/issues/200)
5 tests in the generated test suite needed adjustments in the expectations because of this.

Also I decided to drop json_serializable (see #293). I think it's a good time and it was a lot easier to unify the serialization methods.

I would really appreciate real-world feedback from usage.

Last thing, I plan to migrate document_client in a follow up pull request.


isoos:
You are a hero! I've randomly sampled a few files (in generator, generated and shared), and those look great to me.

> For instance, I choose to make Map and List elements non-nullable. I think it gives a more user-friendly API. But it also mean that we can no longer send and receive null elements in list and map.

I think we can afford bold moves here, esp. if we do major version bumps. Let's see how this play out...

> I would really appreciate real-world feedback from usage.

Let's ship it! I think that is the only feasible way to get that :)

Maybe we can add an extra warning somewhere, or release the versions as a prerelease. But I don't think that everyone will upgrade at the same time, and hopefully we will get the feedback in time to correct for the larger gotchas...

isoos (approved):

Schwusch (approved):
> This was not easy!

You can say that again :) Well done!

> For instance, I choose to make Map and List elements non-nullable.

That will probably be true for the most part. In `aws-sdk-js-v3` the models are defined as [Smithy models](https://awslabs.github.io/smithy/1.0/spec/core/model.htm), where nullability is better defined, although not always super clear.
The model definitions in both v2 and v3 are mirroring the same services, although v2 is basically undocumented and does not contain nullability information, like [v3 does](https://awslabs.github.io/smithy/1.0/spec/core/type-refinement-traits.html#sparse-trait).

In the long run, we probably would need to parse v3 models in order to have correct types. I quickly checked a few v3 models, and they seem to be non-nullable for the most part.

Normally I'd like to see the non-generated code in a separate commit from the generated code, but since it passes the tests we might as well just get something out and take it from there.

Again, great job!

xvrh:
Thanks for your feedbacks. I will merge it now and we will iterate on it.

@Schwusch 
> In the long run, we probably would need to parse v3 models in order to have correct types.

Yes, if this is the direction the `js-sdk` is taking, at some point this will be required. But it looks like a big effort too.

Schwusch:
Mentioning #292 here for visibility.

PR #309, opened 2021-04-25
Migrate document_client to null-safety

xvrh:
I only tried it with simple examples, I'm not a user of advanced features. 

Every feedback are welcome :-)

### DIFF
--- a/document_client/lib/document_client.dart
+++ b/document_client/lib/document_client.dart
@@ -43,11 +43,11 @@ class DocumentClient {
   final DynamoDB dynamoDB;
 
   DocumentClient(
-      {@required String region,
-      DynamoDB dynamoDB,
-      AwsClientCredentials credentials,
-      String endpointUrl,
-      Client client})
+      {required String region,
+      DynamoDB? dynamoDB,
+      AwsClientCredentials? credentials,
+      String? endpointUrl,
+      Client? client})
       : dynamoDB = dynamoDB ??
             DynamoDB(
                 region: region,
@@ -58,13 +58,13 @@ class DocumentClient {
   /// Returns a set of attributes for the item with the given primary key by
   /// delegating to DynamoDB.getItem().
   Future<GetOutput> get({
-    @required String tableName,
-    @required Map<String, dynamic> key,
-    List<String> attributesToGet,
-    bool consistentRead,
-    Map<String, String> expressionAttributeNames,
-    String projectionExpression,
-    ReturnConsumedCapacity returnConsumedCapacity,
+    required String tableName,
+    required Map<String, dynamic> key,
+    List<String>? attributesToGet,
+    bool? consistentRead,
+    Map<String, String>? expressionAttributeNames,
+    String? projectionExpression,
+    ReturnConsumedCapacity? returnConsumedCapacity,
   }) async {
     final getItemOutput = await dynamoDB.getItem(
         key: key.fromJsonToAttributeValue(),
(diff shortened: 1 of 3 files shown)


isoos (approved):
yay!

document_client/pubspec.yaml
```diff
@@ -7,13 +7,19 @@ version: 0.1.0
...
 
 dependencies:
   aws_dynamodb_api: ^0.1.1
   shared_aws_api: ^0.2.1
 
+dependency_overrides:
+  aws_dynamodb_api:
+    path: ../generated/aws_dynamodb_api
+  shared_aws_api:
+    path: ../shared_aws_api
```
Schwusch, line 20: I don't think overrides should be committed, in case we accidentally publish it.
We should probably start using `package:melos` in order to not accidentally commit things like this.

document_client/lib/document_client.dart
```diff
@@ -536,13 +544,13 @@ class UpdateOperation extends Operation {
...
-    this.updateExpression,
-    Map<String, String> expressionAttributeNames,
-    Map<String, dynamic> expressionAttributeValues,
-    ReturnValuesOnConditionCheckFailure returnValuesOnConditionCheckFailure,
+    required Map<String, dynamic> value,
+    required String tableName,
+    required String expression,
+    required this.updateExpression,
+    required Map<String, String> expressionAttributeNames,
+    required Map<String, dynamic> expressionAttributeValues,
```
Schwusch, line 552: Are all these actually required?

document_client/lib/document_client.dart
```diff
@@ -520,14 +527,15 @@ class Operation {
...
-    @required this.value,
-    @required this.tableName,
-    this.expression,
-    this.expressionAttributeNames,
-    this.expressionAttributeValues,
+    required this.value,
+    required this.tableName,
+    required this.expression,
+    required this.expressionAttributeNames,
+    required this.expressionAttributeValues,
```
Schwusch, line 538: Is `required` really needed?

document_client/pubspec.yaml
xvrh: I don't think it is possible to publish a package with `dependency_overrides`. Pub will prevent that. So it is safe to leave the overrides until the publication.
Furthermore if we want the CI to succeed, we need to point to the latest version of those package (which are not published yet).

I have never used `melos` but I like the idea of a tool to help for that.

document_client/lib/document_client.dart
xvrh: `expression` and `updateExpression` seem mandatory. I made the other 2 optionals and nullable.

document_client/lib/document_client.dart
xvrh: `expression` seems mandatory. I made the others optional and nullable.

xvrh:
Thanks for the review @Schwusch . Let me know I you see something else

document_client/pubspec.yaml
Schwusch: > I don't think it is possible to publish a package with dependency_overrides

Okay, that's good atleast. Will it hinder folks that refer to the package through a Git reference?

>Furthermore if we want the CI to succeed, we need to point to the latest version

I know https://github.com/fluttercommunity/plus_plugins has this problem, and I think they solve it with `package:melos`. Until we start using that, I can see your point 👍

Schwusch (approved):

document_client/pubspec.yaml
xvrh: > Will it hinder folks that refer to the package through a Git reference?

I think it won't change anything because there already forced to add their own dependency_overrides.

To depend on this package with Git reference you need to add this in your pubspec.yaml:

```yaml
dependencies:
  document_client:
    git:
      url: https://github.com/agilord/aws_client.git
      path: document_client

dependency_overrides:
  aws_dynamodb_api:
    git:
      url: https://github.com/agilord/aws_client.git
      path: generated/aws_dynamodb_api
  shared_aws_api:
    git:
      url: https://github.com/agilord/aws_client.git
      path: shared_aws_api
```

PR #317, opened 2021-08-11
Update deps in generator.

simonbuchan:
Ran into dependency conflicts in `generator` on a clean checkout (shouldn't the `pubspec.lock` be committed?) so updated everything.

Shockingly, not much needed to be changed for all the major version changes here. Builds cleanly and doesn't generate any diffs in `generated` after `dart run bin/generate.dart generate --download` which seems encouraging, though!

Threw in a fix where the test_all.dart file would be generated with `import 'foo\bar.dart' as ...` on windows.

### DIFF
--- a/generator/lib/download_command.dart
+++ b/generator/lib/download_command.dart
@@ -37,8 +37,8 @@ class DownloadCommand extends Command {
 }
 
 Future<void> _fetchApiDefinitions(String reference) async {
-  final response = await http
-      .get('https://api.github.com/repos/aws/aws-sdk-js/zipball/$reference');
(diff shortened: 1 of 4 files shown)


isoos (approved):

isoos:
thanks!

simonbuchan:
Not sure about that CI failure - is that due to it not using the local `shared_aws_client`?

isoos:
> Not sure about that CI failure - is that due to it not using the local shared_aws_client?

Are you referring to the travis failures? I'm not sure if they are still running properly, we should remove them.

Issue #316, opened 2021-08-09
Design / feature: Unsigned requests and temporary credentials


simonbuchan:
First off, thanks for your work so far on this!

I'm looking at using our current services from Flutter, and I've got two issues so far, both of which are pretty closely related and have a decent, but potentially API breaking, fix.

First I would like to be able to use the generated cognito-idp clients for the *non* admin APIs, that is, sign in, confirm codes, etc. as the existing official amplify APIs are awful and confusing and incompatible with this SDK, but I've got the relevant bits needed for the SRP login factored out into a library that sits on top of just `http.Client` so far - but the types / error handling mismatch kind of sucks. Unfortunately, the current design doesn't support *unsigned* service requests, or not providing *any* credentials (not just letting the service default to `AwsClientCredentials.resolve()`). This is described by the JS SDK api models with `"authtype": "none"` on the `"operations"` item, so it should be pretty easy after allowing not having credentials.

Second, the `cognito-identity` service (and others) can hand out temporary credentials, which the `AwsClientCredentials` class doesn't currently handle. The JS SDK handles this differently between v2 and v3: in v2, the `Credentials` base type has a virtual `get()` that derived types can override to update - if needed - the `accessKeyId` etc. properties which must be called before any signing, while in v3, all credentials parameters to the service are simply replaced with a credentials provider function.

If you like, I could take a stab at a PR that supports these? Any feedback or preferences would be welcome.

simonbuchan:
As an example, my preference for implementing this would look something like:

```dart
// basically like before:
class FooService {
  final AwsClientCredentials credentials;

  FooService({
    ...
    this.credentials = AwsClientCredentials.default,
  });
}

// but now:

class AwsClientCredentials {
  // no fields! would require client changes if they were accessing these, but a relatively straightforward fix and makes their code more likely to be correct.

  // default / compatibility is static credentials
  factory AwsClientCredentials({ String accessKeyId, ... }) => StaticAwsClientCredentials({ accessKeyId: accessKeyId, ... });

  // the default provider fallback chain. Since this class is now immutable, and resolution is deferred, this can be a shared static.
  static AwsClientCredentials default = DefaultAwsClientCredentials();

  // all the specific providers...
  factory AwsClientCredentials.ec2() => ...

  // Actually resolves the current values, including refreshing from the external source if needed.
  // Resolves to null if it's AwsClientCredentials.none()
  abstract Future<StaticAwsClientCredentials?> values();
}
```

[...]

PR #318, opened 2021-08-11
Don't sign operations with authtype == 'none'

simonbuchan:
Partial fix for #316 in that this doesn't let you *not* have credentials yet, it will simply not use them for operations qualified as such. I'm successfully using these generated packages with `AwsClientCredentials(accessKey: 'UNUSED', secretKey: 'UNUSED')`, so at least there's that.

I believe *strictly* this is wrong, as e.g. you can use IAM policy conditions on the otherwise unnecessarily signed request? So I'm happy to leave this as draft until I have a way for the `credentials` parameter to be specified as explicitly unsigned.


isoos:
I like this, but let's ask other ideas too. /cc @Schwusch @xvrh

Schwusch:
I think it's a good idea and LGTM 👍

Schwusch:
Travis tests says:
> [SEVERE] json_serializable:json_serializable on lib/model/error.dart:
Generator cannot target libraries that have not been migrated to null-safety.

Not sure what that's about

simonbuchan:
> Not sure what that's about

I guessed on my last PR #317 that it was that the `shared_aws_client` package is being pulled from `pub.dev`, since that's not null safe yet?

Perhaps CI testing with `generate --dev` would fix that, and other PRs where both the shared and generated packages change together, like this one.

Schwusch:
> the shared_aws_client package is being pulled from pub.dev

Oh right, that's inconvenient. That test is a flawed approach, and is covered by the Github Action we have.
Then I say we merge, and remove the Travis test later :+1:

PR #325, opened 2021-12-16
Generate Step Functions (SFN) service

Schwusch:
Currently, a `CHANGELOG.md` is not created when generating a new service.
I added some logic to compensate for that in the `bump` command.


isoos (approved):
lgtm

Schwusch:
Just tried a little script using the service, and it works :tada:

PR #328, opened 2021-12-27
New way of resolving credentials

Schwusch:
As per #316, I'd like to propose a backwards compatible, extendable and composable way of resolving credentials.

It would perhaps open up for use cases highlighted by @simonbuchan, where temporary credentials need to be refreshed, and/or fetched elsewhere asynchronously. Currently no credentials nor crendential provider has to be supplied when only calling unsigned operations.

I started off creating a utility package `aws_credentials_provider_ini` for resolving ini file assumed role credentials through STS.

I am unsure of the naming, perhaps`aws_credentials_providers` are better. We could collect all kinds of providers there.

Usage examples:

```dart
import 'package:aws_credentials_provider_ini/aws_credentials_provider_ini.dart';

final service1 = AwsService(region: 'eu-west-1', credentialsProvider: fromEnvironment); 
final service2 = AwsService(region: 'eu-west-1', credentialsProvider: fromIni); 

// Chain providers
final service3 = AwsService(
    region: 'eu-west-1',
    credentialsProvider: ({Client? client}) async => 
        await fromEnvironment() ?? await fromIni(client: client),
); 

// Cache and refresh when needed
class CachedCredentialProvider {
  final AwsClientCredentialsProvider credentialsProvider;
  AwsClientCredentials? previousCreds;

  CachedCredentialProvider(this.credentialsProvider);

  Future<AwsClientCredentials?> refreshWhenNeeded({Client? client}) async {
    final expiration = previousCreds?.expiration;

    if (previousCreds == null ||
        expiration != null && DateTime.now().isAfter(expiration)) {
      previousCreds = await credentialsProvider(client: client);
    }

    return previousCreds;
  }
}

final cachedCredentialProvider = CachedCredentialProvider(fromIni);

final service4 = AwsService(
    region: 'eu-west-1',
    credentialsProvider: cachedCredentialProvider.refreshWhenNeeded,
); 
```


aws_credentials_provider_ini/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
+name: aws_credentials_provider_ini
```
xvrh, line 1: Why not use `shared_aws_api` for this code? Is it because of the cyclic dependency on `aws_sts_api`?

Schwusch: >  Is it because of the cyclic dependency on aws_sts_api?

Yup, since STS depends on `shared_aws_api` it broke in extraordinary ways :)

xvrh: What kind of error did you saw exactly? I thought pub handled circular dependencies 💁?

Schwusch: When adding `aws_sts_api` as a dependency to `shared_aws_api`, the analyzer is fine and no error is reported.
When running it though, this error is thrown:
```
Error: Couldn't resolve the package 'aws_sts_api' in 'package:aws_sts_api/sts-2011-06-15.dart'.
../shared_aws_api/lib/src/credentials/credentials_io.dart:3:8: Error: Not found: 'package:aws_sts_api/sts-2011-06-15.dart'
import 'package:aws_sts_api/sts-2011-06-15.dart';
       ^
../shared_aws_api/lib/src/credentials/credentials_io.dart:143:17: Error: Method not found: 'STS'.
    final sts = STS(
```
Then I tried overriding `shared_aws_api` itself on the off chance it could be the circular dependency messing with the compiler:
```yaml
name: shared_aws_api

dependencies:
  aws_sts_api: ^1.0.0

dependency_overrides:
  shared_aws_api:
    path: .
```
That throws this error:
```
[shared_aws_api] dart pub get
Resolving dependencies...
Error on line 25, column 3 of pubspec.yaml: A package may not list itself as a dependency.
   ╷
25 │   shared_aws_api:
   │   ^^^^^^^^^^^^^^
   ╵
```
It was at that time I glanced at `aws-sdk-js-v3` since I know it's modular and would have the same problem.
There they solved it with a separate package like the way I did it here.

PR #333, opened 2022-01-06
Introduce asynchronous credentials

Schwusch:
I'm splitting up #328 since the activity has been low, perhaps because of PR size.
This is introducing the ability to provide an asynchronous credentials provider to the protocol classes, which evaluates only when requests need to be signed.
This improves a number of use cases:
- When there is no need to provide credentials e.g. in
  - aws_sso_oidc_api
  - aws_sso_api
  - aws_cognito_idp_api
  - aws_cognito_identity_api
- When credentials need to be fetched asynchronously, and updated when they expire (e.g. from assumed roles in STS/logged in users in Cognito).
- Provide all the ways to fetch credentials as separate features split from the shared package (due to circular dependency issues):
  - Cognito
  - STS
  - EC2
  - Web Identity
  - SSO

Adding the `credentialsProvider` to the generated service constructors will come in later PRs

### DIFF
--- a/shared_aws_api/lib/src/credentials.dart
+++ b/shared_aws_api/lib/src/credentials.dart
@@ -2,8 +2,12 @@
 // Use of this source code is governed by a BSD-style license that can be found
 // in the LICENSE file.
 
+import 'package:http/http.dart';
 import 'credentials/credentials_io.dart'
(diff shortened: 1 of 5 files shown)


isoos (approved):
lgtm

Schwusch:
Thanks for :eyes:

PR #335, opened 2022-01-06
Migrate Travis tests to Github Actions

Schwusch:
The red cross from Travis is an eye sore, and it's nice to collect everything in one place.
I am not acquainted with Travis/mono_repo enough to know if I missed something, so please speak up if there's more I should remove.


.github/workflows/test.yaml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dart-lang/setup-dart@v1
+      - uses: actions/checkout@v2
+
+      - name: Print Dart SDK version
+        run: dart --version
+
+      - name: Analyze shared_aws_api package
+        run: dart pub get && dart analyze --fatal-warnings --fatal-infos
```
xvrh, line 20: Nit: `--fatal-warnings` is already the default

xvrh (approved):

xvrh:
For the error with json_serializable builder. I think you will need to pin down an older version. Like:

```yaml
  json_annotation: 3.1.1

dev_dependencies:
  analyzer: 1.7.2
  json_serializable: 3.5.1
```

(forcing analyzer version is to workaround a bug I faced recently).

isoos (approved):

.github/workflows/test.yaml
Schwusch: Nice, thanks!

Schwusch:
> For the error with json_serializable builder. I think you will need to pin down an older version. Like:
> 
> ```yaml
> dependencies:
>   json_annotation: 3.1.1
> 
> dev_dependencies:
>   analyzer: 1.7.2
>   json_serializable: 3.5.1
> ```
> 
> (forcing analyzer version is to workaround a bug I faced recently).

Right, after juggling versions back and forth it comes down to pinning `dart_style` to somewhere `>=1.0.0 <2.0.0`, which is a pity.
Perhaps it's better to upgrade the generator to null safety in a later PR.
I'll at least put `continue-on-error: true` on the steps so they'll give more information.

Schwusch:
huh, `continue-on-error: true` marks the step as successful, which wasn't the intent. I'll remove it again.

PR #336, opened 2022-01-06
Add option to pass credentialsProvider in generated services

Schwusch:
This adds the `credentialsProvider` argument to generated services and passes it on to the protocol classes.


isoos (approved):

isoos:
Note: I think we may need to restructure the directory names a bit: `generator/` should come as first in such cases.

Schwusch:
> Note: I think we may need to restructure the directory names a bit: `generator/` should come as first in such cases.

Do you mean when scrolling through changed files? Yeah it's a hassle. I saw @xvrh was experimenting with [`.gitattributes`](https://github.com/xvrh/aws_client/blob/xha/experiment-generated/.gitattributes), maybe that could alleviate some of it.
Otherwise I usually split up the generated files in a separate commit so they can be viewed separately.

PR #337, opened 2022-01-07
Update generator to null safety

Schwusch:
It would be nice to have working tests :)


generator/lib/model/shape.dart
```diff
@@ -150,10 +150,10 @@ class Shape {
     if (type.isBasicType()) {
       return false;
     }
-    if (member != null && !member.shapeClass.requiresJson) {
+    if (!(member?.shapeClass?.requiresJson == true)) {
```
isoos, line 153: Maybe a simpler form?
`if (member?.shapeClass?.requiresJson != true) {`
(same below)

isoos (approved):
lgtm
there may be some opportunity to help the compiler make earlier decisions about non-nullability, but this looks good, and the micro-optimizations can be done if/when we are touching that part of the code.

generator/lib/model/shape.dart
Schwusch: good eye :+1:

Schwusch:
> there may be some opportunity to help the compiler make earlier decisions about non-nullability, 

Yeah, this was a least effort migration, and I tried to make the amount of changes small.
There's a little more exclamation marks than I'm usually comfortable with.

Schwusch:
This reminds me that we don't have CI tests checking formatting. I'll add that as well.

PR #341, opened 2022-01-08
Run publish operations in parallel

Schwusch:
I got tired of waiting when publishing, and this made publishing 12 times faster on my machine, which is very appreciated :)

### DIFF
--- a/generator/lib/package_command.dart
+++ b/generator/lib/package_command.dart
@@ -3,6 +3,8 @@ import 'dart:io';
 
 import 'package:args/command_runner.dart';
 import 'package:http/http.dart';
+import 'package:pedantic/pedantic.dart';
+import 'package:pool/pool.dart';
 import 'package:version/version.dart';
 import 'package:yaml/yaml.dart';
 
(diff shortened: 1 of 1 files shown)


generator/lib/package_command.dart
```diff
@@ -238,52 +240,56 @@ class PublishCommand extends Command {
     final packages = await _scanPackages(config);
 
     final client = Client();
+    final pool = Pool(Platform.numberOfProcessors);
```
isoos, line 243: Let's use a reasonable max (e.g. 4 or 8).

generator/lib/package_command.dart
```diff
@@ -238,52 +240,56 @@ class PublishCommand extends Command {
...
+              .join('\n'));
+          if (pr.exitCode != 0) {
+            throw Exception('Unable to publish.');
+          }
         }
-      }
+      }));
     }
 
+    await pool.close();
```
isoos, line 292: Oh, I didn't know that you don't need to wait for the futures to complete.

isoos (approved):

generator/lib/package_command.dart
Schwusch: I guess it awaits everything in the queue to be processed before closing?
I just copied what was done in the generate command.

generator/lib/package_command.dart
Schwusch: How come 4 or 8? And what makes it reasonable?
This is taken from generate command, it seems to work? 乁( •_• )ㄏ

PR #344, opened 2022-01-24
Resolve credentials from AWS IDMS

dsyrstad:
For use when running on EC2 instances. It is the last fallback in the chain. Local credentials are attempted first.

### DIFF
--- a/shared_aws_api/lib/src/credentials.dart
+++ b/shared_aws_api/lib/src/credentials.dart
@@ -39,5 +39,6 @@ class AwsClientCredentials {
     this.expiration,
   });
 
-  static AwsClientCredentials? resolve() => CredentialsUtil.resolve();
+  static Future<AwsClientCredentials?> resolve(Client? client) =>
+      CredentialsUtil.resolve(client);
 }
--- a/shared_aws_api/lib/src/credentials/credentials_io.dart
+++ b/shared_aws_api/lib/src/credentials/credentials_io.dart
@@ -1,11 +1,18 @@
+import 'dart:convert';
 import 'dart:io';
 
+import 'package:http/http.dart';
 import 'package:shared_aws_api/src/credentials.dart';
 import 'package:shared_aws_api/src/credentials/ini_config.dart';
 
 class CredentialsUtil {
-  static AwsClientCredentials? resolve() {
-    return fromEnvironment ?? fromProfileFile;
+  static Future<AwsClientCredentials?> resolve(Client? client) {
+    final localCredentials = fromEnvironment ?? fromProfileFile;
+    if (localCredentials != null) {
+      return Future.value(localCredentials);
+    }
+
+    return fromInstanceMetaDataService(client);
   }
 
   static AwsClientCredentials? get fromEnvironment {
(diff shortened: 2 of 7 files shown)


dsyrstad:
Ping?

shared_aws_api/lib/src/credentials.dart
```diff
@@ -39,5 +39,6 @@ class AwsClientCredentials {
     this.expiration,
   });
 
-  static AwsClientCredentials? resolve() => CredentialsUtil.resolve();
+  static Future<AwsClientCredentials?> resolve(Client? client) =>
```
isoos, line 42: Isn't this a breaking change? Can we do it in an incremental way where we don't break current users?

shared_aws_api/lib/src/credentials/credentials_io.dart
```diff
@@ -1,11 +1,18 @@
...
 import 'dart:io';
 
+import 'package:http/http.dart';
 import 'package:shared_aws_api/src/credentials.dart';
 import 'package:shared_aws_api/src/credentials/ini_config.dart';
 
 class CredentialsUtil {
-  static AwsClientCredentials? resolve() {
-    return fromEnvironment ?? fromProfileFile;
+  static Future<AwsClientCredentials?> resolve(Client? client) {
```
isoos, line 9: Same here: breaking change.

shared_aws_api/lib/src/credentials/credentials_io.dart
```diff
@@ -71,4 +78,53 @@ class CredentialsUtil {
 
     return AwsClientCredentials(accessKey: accessKey, secretKey: secretKey);
   }
+
+  static const _imdsV1BaseSecurityCredentialsUrl =
+      'http://169.254.169.254/latest/meta-data/iam/security-credentials/';
```
isoos, line 83: What is this magic IP? Not even https?

isoos (changes requested):

shared_aws_api/lib/src/credentials/credentials_io.dart
dsyrstad: Yep, it's a magic AWS IP. It's in a private IP range so it won't reach out to the internet. See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials. Also see AWS's own JS SDK: https://github.com/aws/aws-sdk-js/blob/master/lib/metadata_service/get_endpoint.js

shared_aws_api/lib/src/credentials.dart
dsyrstad: I don't think it shouldn't be. It was only being called by the `protocol/` files json.dart,  query.dart, rest-json.dart, and rest-xml.dart - all in the same fashion. In the end, they all use it to create an `AwsClientCredentialsProvider`, which returns a `Future`:

```dart
    } else {
      credentialsProvider ??=
          ({Client? client}) => AwsClientCredentials.resolve(client);
    }
```

I refactored all of the protocol files like so:

```dart
+ ({Client? client}) => Future.value(AwsClientCredentials.resolve());
- ({Client? client}) => AwsClientCredentials.resolve(client);
```

The result is the same for users of the protocols.

shared_aws_api/lib/src/credentials/credentials_io.dart
dsyrstad: See above.

Schwusch:
Thanks @dsyrstad for the initiative!
I'd like to consider `AwsClientCredentials.resolve()` deprecated.
I think new ways of resolving credentials (especially asynchronously) would be better suited as an `AwsClientCredentialsProvider`.
I merged #343, which adds `package:aws_credential_providers` where I would say this would fit nicely.

The only reason `AwsClientCredentials.resolve()` hasn't been removed completely is backwards compatibility and lack of time for replacing it with a sane set of default providers.

dsyrstad:
@Schwusch I don't see `aws_credential_providers` in pub.dev yet. When will you be publishing it? Also , `aws_credential_providers`  doesn't seem to support a chained fallback mechanism - e.g., try to resolve from env vars, if that fails,  then try the ini file/AWS_PROFILE, then IMDS  (my case). In this PR, it falls back gracefully as I just described. The chained fallback mechanism would allow the same code to run in different environments (local machine, EC2, ECS, etc.)

If you haven't done so already, I'd suggest looking at the AWS JS SDK and modeling the provider implementation on that. The Java SDK also works in a similar manner.

Schwusch:
> When will you be publishing it?

It's a work in progress :) And as you mention, it does not have all the features you'd want, like falling back gracefully through a "provider chain".
You're most welcome to fix that as you see fit, and when your use-case is covered, I think it's worth publishing.
After that, my plan is to add a default provider chain to all services, which adjusts/is compatible with both web and elsewhere. 
Everything done here is best effort, so it may not be an exact copy of official AWS SDKs approach.

The implementation I had in my head, regarding the chained fallback mechanism, goes something like this:
```dart
class ChainedCredentialProvider {
  final List<AwsClientCredentialsProvider> credentialProviders;
  static const defaultChain = [
    fromEnvironment,
    fromIni,
    // TODO: AWS IDMS
  ];

  ChainedCredentialProvider(this.credentialProviders);

  Future<AwsClientCredentials?> chainedFallback({Client? client}) async {
    for (final provider in credentialProviders) {
      final credentials = await provider(client: client);
      if (credentials != null) {
        return credentials;
      }
    }

    return null;
  }

  static Future<AwsClientCredentials?> defaultProviderChain({Client? client}) async {
    for (final provider in defaultChain) {
      final credentials = await provider(client: client);
      if (credentials != null) {
        return credentials;
      }
    }

    return null;
  }
}
```
Then usage becomes:
```dart 
final sqs = SQS(credentialsProvider: ChainedCredentialProvider.defaultProviderChain);
```
And later, I'm thinking the SQS constructor should look something like 
```dart
SQS({
    String? region,
    _s.AwsClientCredentials? credentials,
    _s.AwsClientCredentialsProvider credentialsProvider = ChainedCredentialProvider.defaultProviderChain,
    _s.Client? client,
  })
```
You're most welcome to submit a PR, if not I can probably do it when I have some spare time over at some point.

dsyrstad:
@Schwusch Thanks for your response. I'll let you know if I start working on it, I'm pretty busy at the moment.  If you start before me, please let me know.

Schwusch:
Since #351 is merged and the [aws_credential_providers](https://pub.dev/packages/aws_credential_providers#links) is published, is it OK to close this?

dsyrstad:
> Since #351 is merged and the [aws_credential_providers](https://pub.dev/packages/aws_credential_providers#links) is published, is it OK to close this?

Yes, please.

Schwusch:
If you use `aws_credential_providers` and find a bug, you're most welcome to open a new issue or PR :)

PR #349, opened 2022-02-02
Update README files

Schwusch:
I'd like to suggest some updates to all READMEs:
- Some more info on the main readme.
- Remove Travis/mono_repo stuff, since it's not used anymore
- Adjust generator readme to current functionality
- A direct link to the main class API reference in the generated service readme
- Remove warnings and contributors in generated readmes, we're more than just 2 contributors now and warnings don't add value at this point.

If there is more that I should add, give a commment.


isoos (approved):
I'd keep the warning about being a generated library, but maybe de-emphasize it by moving to the end? Other than that, lgtm!

PR #350, opened 2022-02-13
Organize code a bit in aws_credential_providers

Schwusch:
With #344 in mind, it should perhaps be a little more organized and clear how to add new credential providers.
This is mostly splitting functionalities up in separate files/folders.

@dsyrstad I can't add you as a reviewer, but I'd like to hear your input if you have any.


dsyrstad:
LGTM. I guess I'll know how well it works once I try to implement the IMDS provider.

PR #352, opened 2022-02-18
Fix IMDS provider signature, make web compatible

Schwusch:
Taking another look at the credential providers, there were type errors needed fixing.
I'm looking at putting `defaultProvider` as a default value for the services constructors `credentialsProvider` , effectively replacing `AwsClientCredentials.resolve()`.
That requires the package to be web compatible, so that's fixed too.

cc @dsyrstad


dsyrstad:
OMG - someone actually uses AWS credentials from a web app?? LGTM

PR #359, opened 2022-06-11
Miscellaneous Fixes

mohsinhijazee:
Fixes following things:

### In Shared API 

- The `extractXmlChild()` returns the element itself if the name of the element matches with the name given as parameter.

Without it, the calls such as [GetBucketLocation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html) fail as they have the root element as their output:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<LocationConstraint xmlns="http://s3.amazonaws.com/doc/2006-03-01/">eu-central-1</LocationConstraint>
```         

### In generator

- The dart extensions are named so that they are publicly visible as well. Without it, the `toValue()` and `toSomeEnum()` on the String do not work.
- The `isdevMode` in package manifest building simply adds the `shared_aws_api` as path directly to the `dependencies` as well as `dependency_overrides`. This is done because when depending on some specific package (such as `aws_s3_api`) from an app, by pointing to the local copy of this package still fetches the `shared_aws_api` from the dart package registry even if the `dependency_overrides` stanza is present.

All generated packages are regenerated after the above changes.

FYI @isoos happy to make adjustments.

### DIFF
--- a/generated/aws_accessanalyzer_api/lib/accessanalyzer-2019-11-01.dart
+++ b/generated/aws_accessanalyzer_api/lib/accessanalyzer-2019-11-01.dart
@@ -952,7 +952,7 @@ enum AnalyzerStatus {
   failed,
 }
 
-extension on AnalyzerStatus {
+extension AnalyzerStatusValue on AnalyzerStatus {
   String toValue() {
     switch (this) {
       case AnalyzerStatus.active:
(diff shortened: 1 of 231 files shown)


isoos:
I think this looks good.
/cc @Schwusch @xvrh - any concern about the extensions becoming public?

Schwusch:
LGTM, I suspect the extensions won't bother anyone :)

mohsinhijazee:
I have few typos + some analyze/linting erros (which I am not much familar with), would push the changes as I fix them.

PR #360, opened 2022-07-31
Modify to specify endpoint value in constructor for endpoint custom

paca94:
hi, I am using it to personally create and use a dashboard for localstack.

While implementing `SQS`, it was not possible to specify and use `endpointUrl`, so the corresponding part was modified.


Schwusch:
Nice, thanks!

Clean PR, in it goes!

Issue #361, opened 2022-07-31
Class 'QueueAttributeName' has no instance method 'toJson'.


paca94:
hello. If you create a queue using the code below, an error occurs.
```dart
createQueue(
            queueName: "testQueue",
            attributes: {QueueAttributeName.fifoQueue: "true"})
```

The error that occurs is as follows.
```
[ERROR:flutter/lib/ui/ui_dart_state.cc(198)] Unhandled Exception: NoSuchMethodError: Class 'QueueAttributeName' has no instance method 'toJson'.
Receiver: Instance of 'QueueAttributeName'
Tried calling: toJson()
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
#1      _flatten (package:shared_aws_api/src/protocol/query.dart:253:17)
#2      _SyncIterator.moveNext (dart:core-patch/core_patch.dart:186:26)
#3      MapMixin.addEntries (dart:collection/maps.dart:177:23)
#4      new Map.fromEntries (dart:core/map.dart:209:17)
#5      flatQueryParams (package:shared_aws_api/src/protocol/query.dart:163:14)
#6      QueryProtocol._buildRequest (package:shared_aws_api/src/protocol/query.dart:122:9)
#7      QueryProtocol.send (package:shared_aws_api/src/protocol/query.dart:75:22)
#8      SQS.createQueue (package:aws_sqs_api/sqs-2012-11-05.dart:748:37)
#9      Sqs.buildCreateQueueButton.<anonymous closure> (package:localstack_dashboard_client/src/screens/sqs.dart:126:26)
<asynchronous suspension>
```

Schwusch:
Hmm, seems `QueueAttributeName` isn't registered int the generator as an attribute that needs a `.toJson()` generated.
Ping @xvrh , was your custom serialisation merged or are we using json_serializable these days? 😅

xvrh:
@Schwusch if I remember well it is indeed my (😇 ) custom serialisation at this point. It seems map keys are not well handled.
I'll have a look now.

xvrh:
Not sure if the json serialiser was really at fault as it comes from the Query protocol. I submitted tentative fix.

PR #362, opened 2022-08-01
Handle enum in map in the Query protocol

xvrh:
Fixes: https://github.com/agilord/aws_client/issues/361 (hopefully)

I added a unit test but I didn't test with the real SQS service. I would appreciate if someone (@paca94 @Schwusch) can smoke test it.


xvrh:
I tested it with real SQS service and I can create a queue with this code:

```dart
  final sqs = Sqs(region: 'us-east-1');
  final queue = await sqs.createQueue(
      queueName: 'my-queue', attributes: {QueueAttributeName.delaySeconds: '5'});
```

PR #364, opened 2022-08-24
Include x-amz-security-token in set of signed headers

dmosdallas:
When sending a request using this signing scheme from ECS tasks on fargate, I get the following response:

```xml
<Error><Code>AccessDenied</Code><Message>There were headers present in the request which were not signed</Message><HeadersNotSigned>x-amz-security-token</HeadersNotSigned>
```

Moving the addition of the `X-Amz-Security-Token` header before the canonicalization of headers resolved this error for me.


Schwusch:
Thank you @dmosdallas for your help! I'll make a new release of the shared package

PR #368, opened 2022-11-15
Marshall SS (string set) type

bschulz87:
This PR adds SS (string set) type to the DynamoDB Document Client translator.


Schwusch:
Thank you @bschulz87 for the PR! A few things:
The conversion tests fail, it expects a list with strings, see below. I think it would be more correct to convert `Set<String>` to a string set? 
Not sure how the users would create a list of strings otherwise. 
Great initiative 👍 
```
test/conversions_test.dart: Converting convert back and forth between JSON and DynamoDB Input/Output (failed)
  Expected: {
              'foo': {'BOOL': true},
              'bar': {'S': 'baz'},
              'fizz': {
                'M': {
                  'buzz1': {'B': 'AA=='},
                  'buzz2': {'L': [{'S': 'stringyDingy'}]},
                  'buzz3': {'BS': ['AA==']},
                  'buzz4': {'NS': ['1', '2', '3']}
                }
              },
              'noCreativity': {'NULL': true}
            }
    Actual: {
              'foo': {'BOOL': true},
              'bar': {'S': 'baz'},
              'fizz': {
                'M': {
                  'buzz1': {'B': 'AA=='},
                  'buzz2': {'SS': ['stringyDingy']},
                  'buzz3': {'BS': ['AA==']},
                  'buzz4': {'NS': ['1', '2', '3']}
                }
              },
              'noCreativity': {'NULL': true}
            }
     Which: at location ['fizz']['M']['buzz2'] is {'SS': ['stringyDingy']} which is missing map key 'L'
     ```

bschulz87:
Thank you for your feedback. I updated the translator and also extended the test to expect a proper "SS" translation back from the output!

Schwusch:
Nice work! Now it's only complaining about format, easily fixed with a `dart format .` :)

Schwusch:
Let's bring it in!
Well done

Schwusch:
This is now published as [2.0.0](https://pub.dev/packages/document_client)

PR #374, opened 2022-12-06
Actually use calculateDateTimeToJson when generating services

### LINKED ISSUE #371: MalformedInput Sender: timestamp must follow ISO8601
When using [`CloudWatch.putMetricData()`](https://github.com/agilord/aws_client/blob/e25e160b4912af146d7ce173fcf186f0489ee5c7/generated/aws_cloudwatch_api/lib/monitoring-2010-08-01.dart#L2589), this error is observed:

```
MalformedInput Sender: timestamp must follow ISO8601
#0 QueryProtocol.send (package:shared_aws_api/src/protocol/query.dart:100:7)
<asynchronous suspension>
#1 CloudWatch.putMetricData (package:aws_cloudwatch_api/monitoring-2010-08-01.dart:2605:5)
<asynchronous suspension>
#2 FutureExtensions.onError.<anonymous closure> (dart:async/future.dart:1014:15)
<asynchronous suspension>
```

Schwusch:
Hopefully fixes #371 :pray: 
Ping @cbatson

### DIFF
--- a/generator/lib/builders/builder_utils.dart
+++ b/generator/lib/builders/builder_utils.dart
@@ -2,6 +2,7 @@ import 'package:aws_client.generator/model/descriptor.dart';
 import '../model/dart_type.dart';
 import '../model/shape.dart';
 import '../utils/string_utils.dart';
+import 'library_builder.dart';
 
 class Nullability {
   static final inputOutput = Nullability._(true, true);
@@ -118,8 +119,9 @@ String encodeJsonCode(Shape shape, String variable,
     shape.isTopLevelInputEnum = true;
     return '$variable${nullability.inputNullAware}.toValue()${nullability == Nullability.input ? "?? ''" : ''}';
   } else if (shape.type == 'timestamp') {
-    final timestampFormat =
-        member?.timestampFormat ?? shape.timestampFormat ?? 'unixTimestamp';
+    final timestampFormat = member != null
+        ? calculateDateTimeToJson(member)
+        : shape.timestampFormat ?? 'unixTimestamp';
     return '${timestampFormat}ToJson($variable)';
   } else if (shape.type == 'blob') {
     if (nullability.inputNullable) {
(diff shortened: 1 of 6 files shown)


generator/lib/builders/builder_utils.dart
```diff
@@ -118,8 +119,9 @@ String encodeJsonCode(Shape shape, String variable,
     shape.isTopLevelInputEnum = true;
     return '$variable${nullability.inputNullAware}.toValue()${nullability == Nullability.input ? "?? ''" : ''}';
   } else if (shape.type == 'timestamp') {
-    final timestampFormat =
-        member?.timestampFormat ?? shape.timestampFormat ?? 'unixTimestamp';
+    final timestampFormat = member != null
+        ? calculateDateTimeToJson(member)
+        : shape.timestampFormat ?? 'unixTimestamp';
```
xvrh, line 124: Shouldn't `shape.timestampFormat` have priority here?

```dart
final timestampFormat = shape.timestampFormat ?? (member != null
        ? calculateDateTimeToJson(member)
        :  'unixTimestamp');
```

Maybe it doesn't matter...

xvrh (approved):
Nice!

generator/lib/builders/builder_utils.dart
Schwusch: Not sure of the order, but it is translated from 
https://github.com/aws/aws-sdk-js/blob/master/lib/model/shape.js#L283-L286 to
https://github.com/agilord/aws_client/blob/00d8f2da0107e0cd83926546abead0217d8c3f2d/generator/lib/builders/library_builder.dart#L411-L414

It could be the other way around, not sure how members and shapes translates in JS to our implementation.

Schwusch: I would guess it's pretty rare for a member and its shape to have conflicting timestamp formats.

cbatson:
Nice, thank you for this!! The logic looks correct to me, or at least looks to have parity with the JS approach.

For the generated files in this diff, I checked the docs to verify the timestamp formats. All appear to match except for the `lastAttemptDate` field of `RecipientDsnFields` which [according to the docs](https://docs.aws.amazon.com/ses/latest/APIReference/API_RecipientDsnFields.html) should be RFC 822. I don't think that should delay this PR, since the existing coding of it as a unix timestamp is almost certainly incorrect. But it might be worth at some point digging to see how it's treated in JS to see if there's any further disparity in the Dart translation.

Also, if it's ok with you, I'd like to add a test specifically for the CloudWatch `MetricDatum` to catch future regressions. I can put up a PR against this branch.

Issue #241, opened 2020-09-13
Tackle signing protocols


Schwusch:
It seems we have overlooked signing protocols a bit, as showed in #238, and I think it's time to get an overview of what we need to support.

This issue is for tracking the work towards supporting the missing signing protocols.

jcblancomartinez:
Hi @Schwusch,

I'm also having the same issue as @littlekid440 

When signing S3's PutObject operation via aws_s3_api, I'm getting:

<Error><Code>AccessDenied</Code><Message>There were headers present in the request which were not signed</Message>
<HeadersNotSigned>x-amz-security-token</HeadersNotSigned>

It seems that S3 expects 'X-Amz-Security-Token' header to be signed in and here you are not signing it
https://github.com/agilord/aws_client/blob/master/shared_aws_api/lib/src/protocol/_sign.dart#L65-L67

Do you have any ETA on when this will be addressed?

Thanks.

Schwusch:
S3 seems to be a little bit special and is tracked in #238

Schwusch:
As mentioned in #238, I think I have a way forward with this.
The trick is to expose signAws4HmacSha256 and other signing functions in the shared_aws_api, and let the protocol class send method take an optional signing function as an argument. The function signature would be something like:
```dart
typedef RequestSigner = void Function({
  required Request rq,
  required ServiceMetadata service,
  required String region,
  required AwsClientCredentials credentials,
});
```
E.g. RestXmlProtocol.send() method signature would then be:
```dart
Future<RestXmlResponse> send({
    required String method,
    required String requestUri,
    required Map<String, AwsExceptionFn> exceptionFnMap,
    bool signed = true,
    Map<String, List<String>>? queryParams,
    Map<String, String>? headers,
    dynamic payload,
    String? resultWrapper,
    RequestSigner? requestSigner, // take signer as an argument
  }) async {}
```
Later, the signing could simply be:
```dart
if (requestSigner != null) {
  requestSigner(
    rq: rq,
    service: _endpoint.service,
    region: _endpoint.signingRegion,
    credentials: credentials,
  );
} else {
  signAws4HmacSha256(
    rq: rq,
    service: _endpoint.service,
    region: _endpoint.signingRegion,
    credentials: credentials,
  );
}
```
That way we only have to supply the send() method a signing function when it's not supposed to be V4 signing.
During service code generation, it's possible to then decide what signing function (if any!) the operation needs, and provide it.

ryanheise:
[...]

PR #378, opened 2022-12-07
Bump packages for release

Schwusch:
Not sure how much is breaking (e.g. package:xml is bumped a major version), so I took the liberty to bump a major version.


xvrh (approved):
👍

Schwusch:
All packages are released as 2.0.0 now @cbatson

cbatson:
Nice!! I've been able to go back to the published packages for all but S3 now. Still dealing with #241, for which I have a local edit. Perhaps I will turn it into something reasonable and PR it.

PR #403, opened 2023-06-25
Remove double-encode of request URI when generating signature

### LINKED ISSUE #402: s3 getObject `SignatureDoesNotMatch` if path contains a `:`
Easy repro — upload a file named `test:file.json` with any contents and attempt to fetch it using s3.getObject.

Request will fail with `SignatureDoesNotMatch`.

I'm going to look into this briefly to see if I can find out why `:` breaks signing, but don't claim to be an AWS expert. Any experienced S3 people have any pointers/ideas on this one?

VerTiGoEtrex:
Fixes #402


xvrh (approved):

xvrh:
Thanks for the contribution!

PR #419, opened 2024-03-14
Fix Payload root element name

dreznichuk:
Hi
YetAnotherCompleteMultipartUpload fix

But in this case it is fixed in generator code, not in generated like in https://github.com/agilord/aws_client/pull/413

All other commits were auto-generated

The only place where needed element name were in api schema is locationName from payload memeber

As you can see it turned out that `CompleteMultiPartUpload` was the only place with naming mismatch


Schwusch:
Looking good 👍 Nice work with actually generating the code

PR #425, opened 2024-06-08
Refactor query protocol

xvrh:
The query protocol was different than the other protocols (rest, xml). 
The serialization format is stored in a model (thin model embedded as a json map literal). And at runtime, it uses that model to perform the serialization.

The other protocols, have the serialization rules directly generated as Dart code.

I refactored the protocol to get rid of the thin model and generate the serialization rules in code.

The main advantage is that it's better for tree-shaking and more inline with the other protocols.


isoos (approved):
lgtm

PR #426, opened 2024-06-09
Refactor enum handing

xvrh:
Instead of generating 2 extensions with `toValue()` and `to<EnumName>()`, we use "enhanced enums" feature to store the `value` string directly in the enum.


isoos (approved):
lgtm

isoos:
Note: I think I'll deprecate most of the old, per-API packages in the coming days/weeks on pub.dev. After that, we may just delete these in the repository, to reduce the overhead of such PRs.

xvrh:
Thanks

> I think I'll deprecate most of the old, per-API packages in the coming days/weeks on pub.dev

@isoos ok, let me know when this is done so I can clean up the repo & generator appropriately.

PR #451, opened 2025-07-10
Fix missing enum value in LambdaUpdateStatusReasonCode

xvrh:
This is missing from the original schema. I hardcode a fix in the generator


xvrh:
I'm going ahead and publish this

isoos (commented):
lgtm

PR #456, opened 2026-05-06
Update pubpec.yaml for `shared_aws_api` to allow `intl` up to and including 0.20.x

ralmand-gridware:
Currently `shared_aws_api` is updated for `intl` v0.20.x, however the `pubspec.yaml` restricts intl to `>=0.18.0 <0.20.0`, preventing dependency resolution for anyone using `intl` at v0.20.0 or greater. `intl` v0.20.0 was released in November 2024, which causes dependency resolution problems for recently created Dart and Flutter projects.


ralmand-gridware:
Fixed upstream.

PR #460, opened 2026-06-05
Remove multi-package support, keep only the aws_client package

xvrh:
Collapse the repo to a single published package (aws_client) ahead of a larger generator change (migration to Smithy).

Packages removed (generated/, all_apis/, shared_aws_api/, document_client/, aws_credential_providers/). The shared runtime, DynamoDB document client and credential providers now live as hand-maintained source under aws_client/lib/src/.


isoos:
image
:100:

isoos (approved):

PR #468, opened 2026-06-16
Use Smithy models

xvrh:
Implies some discontinued services dropped, many new ones added. ~2 years of model drift to catch up


isoos (approved):
thanks for all this work!

